PR #5, opened 2022-07-26
update from other repo

cqc-melf:


cqc-melf:
The PR includes all the updates from the old pytket-extension repo

cqc-alec (approved):
Looks good.
@ss2165 should we include https://github.com/CQCL/pytket-extensions/pull/454 before releasing?

cqc-melf:
I was not sure why the PR is not merged yet and for that reason I have not included it here already. If we want it to be included I am happy to add that.

PR #10, opened 2022-07-26
Release/0.6.1

cqc-melf:
fix doc build and update version for new release


cqc-melf:
I have not added something to the changelog on purpose, if you think it would be better to add that I am happy to do that. @cqc-alec

cqc-alec:
Do you want to add a line to the changelog?

cqc-alec (approved):

PR #15, opened 2022-07-29
fix link to examples

cqc-melf:


cqc-melf:
We should think about moving the examples to the pytket repo, this is not a good placed for them. We still should fix the link issue for now.

cqc-melf:
And I have changed the link to pointing to the main branch, which avoids this kind of issues.

cqc-alec (approved):

PR #17, opened 2022-08-02
add offline api

cqc-melf:


pytket/extensions/quantinuum/backends/api_wrappers.py
```diff
@@ -469,3 +470,130 @@ def cancel(self, job_id: str) -> dict:
...
+        ],
+    ):
+        """Initialize OfflineQuantinuum API client.
+        Trys to allow all the operations like the QuantinuumAPI without
+        any interaction with the remote device
+        All jobs that are submitted to this offline api are stored
+        and can be requested again later
+
+        :param machine_list: gives the parameter of all the devices
+        and the device information
```
cqc-alec, line 514:
```suggestion
        """Initialize offline API client.
        
        Tries to allow all the operations of the QuantinuumAPI without
        any interaction with the remote device.
        
        All jobs that are submitted to this offline API are stored
        and can be requested again later.

        :param machine_list: gives the parameter of all the devices
            and the device information
```

.pylintrc
```diff
@@ -13,7 +13,6 @@ enable=
     bad-super-call,
     consider-merging-isinstance,
     continue-in-finally,
-    dangerous-default-value,
```
cqc-alec, line 16: I think this is useful to keep.

pytket/extensions/quantinuum/backends/api_wrappers.py
```diff
@@ -469,3 +470,130 @@ def cancel(self, job_id: str) -> dict:
         jr = res.json()
 
         return jr  # type: ignore
+
+
+class QuantinuumAPIOffline:
+    def __init__(
+        self,
+        machine_list: Optional[list] = [
```
cqc-alec, line 478:
```suggestion
        machine_list: list = [
```
No need for `Optional` I think?

pytket/extensions/quantinuum/backends/api_wrappers.py
```diff
@@ -469,3 +470,130 @@ def cancel(self, job_id: str) -> dict:
...
+            },
+        ],
+    ):
+        """Initialize OfflineQuantinuum API client.
+        Trys to allow all the operations like the QuantinuumAPI without
+        any interaction with the remote device
+        All jobs that are submitted to this offline api are stored
+        and can be requested again later
+
+        :param machine_list: gives the parameter of all the devices
```
cqc-alec, line 513: To be useful, this docstring should explain the format required, or at least give an example.

tests/backend_test.py
```diff
@@ -104,6 +105,56 @@ def test_quantinuum(
...
+    c.measure_all()
+    c = backend.get_compiled_circuit(c)
+    n_shots = 4
+    _ = backend.process_circuits([c], n_shots)[0]
+    expected_result = {
+        "name": "test 1",
+        "count": 4,
+        "machine": "H1-1",
+        "language": "OPENQASM 2.0",
+        "program": """OPENQASM 2.0;\ninclude "hqslib1.inc";\n\nqreg q[4];\ncreg c[4];
```
cqc-alec, line 131: Hard-coding the result of the compilation like this seems very fragile: details of compilation passes could change and teh API doesn't guarantee a particular QASM string. I don't think we should assert equality here.

pytket/extensions/quantinuum/backends/api_wrappers.py
```diff
@@ -469,3 +470,130 @@ def cancel(self, job_id: str) -> dict:
         jr = res.json()
 
         return jr  # type: ignore
+
+
+class QuantinuumAPIOffline:
```
cqc-alec, line 475: Currently the `_api_handler` argument to the backend constructor is underscored. Now that we have a (public, documented) alternative, we should probably remove the underscore -- otherwise no-one will know how to use this.

cqc-alec (changes requested):

.pylintrc
cqc-melf: Is there a way to ignore one specific line where this error is occurring? I have been hit by this `dangerous-default-value` without being able to see which parameter it is corresponding to or referring to. Will do some further investigations to find out if there is a different way to resolve this problem.

pytket/extensions/quantinuum/backends/api_wrappers.py
cqc-melf: Mypy wanted it this way, might be a chain of other conditions, will try to find a different way.

tests/backend_test.py
cqc-melf: I have only included this part in the test for other people to see how it would look like, the qasm part is not checked in the test.

pytket/extensions/quantinuum/backends/api_wrappers.py
cqc-melf: Will do, I think the old api handler need some documentation update, too. Will add that as well.

.pylintrc
cqc-melf: Have solved this now.

cqc-melf:
Have added all the documentation for the two apihandler

tests/backend_test.py
cqc-alec: Ah OK, sorry I missed that.

docs/changelog.rst
```diff
@@ -1,6 +1,13 @@
 Changelog
 ~~~~~~~~~
 
+0.7.0 (unreleased)
+------------------
+
+* Add new `QuantinuumAPIOffline` for allowing usage of the backend without API calls.
+* New `api_handler` parameter for `QuantinuumBackend`, allowing to choose
+  online or offline options. Default value is the old online api.
```
cqc-alec, line 9:
```suggestion
  online or offline options. Default value is the standard online api.
```

pytket/extensions/quantinuum/backends/api_wrappers.py
```diff
@@ -469,3 +476,143 @@ def cancel(self, job_id: str) -> dict:
...
+
+        Tries to allow all the operations of the QuantinuumAPI without
+        any interaction with the remote device.
+
+        All jobs that are submitted to this offline API are stored
+        and can be requested again later.
+
+        :param machine_list: gives the parameter of all the devices
+            and the device information, the format of this list should be in
+            the same format like the backend would return this. One short example:
```
cqc-alec, line 497:
```suggestion
        :param machine_list: List of dictionaries each containing device information.
            The format of should match what a real backend would return. One short example:
```

cqc-alec (changes requested):
Just a couple of wording suggestions.

cqc-melf:
Thank you for the suggestions, I have committed them!

cqc-alec (approved):

pytket/extensions/quantinuum/backends/api_wrappers.py
dlucchetti: I think the real answer is Optional[list] = None?

If you give it a non-optional value I think you get that as the default?

pytket/extensions/quantinuum/backends/api_wrappers.py
dlucchetti: Maybe the current default parameter could just be moved to the docstring as an example?

pytket/extensions/quantinuum/backends/api_wrappers.py
cqc-melf: If you construct the Quantinuum backend you can decide if you want to hand over the QuantinuumAPI (default) or the new QuantinuumAPIOffline which removes the calls to the remote API. It is currently not possible to switch between them within a created backend, is that something you would like to do?

pytket/extensions/quantinuum/backends/api_wrappers.py
cqc-melf: The QuantinuumAPIOffline will not work without this information, this will be set in case the list given is None. I have added a shorter example in the docstring to explain that a bit more.

cqc-melf:
I will merge this now, but that should not stop any questions from been asked or answered @dlucchetti

PR #19, opened 2022-08-04
add doc build to ci pr checks

cqc-melf:


cqc-melf:
I would like to add this for all extension repos, what do you think? @cqc-alec

.github/workflows/build_and_test.yml
```diff
@@ -70,6 +70,18 @@ jobs:
...
+      if:  (matrix.os == 'ubuntu-20.04') && (github.event_name == 'pull_request')
+      run: |
+        pip install -r .github/workflows/docs/requirements.txt
+    - name: Build docs
+      if:  (matrix.os == 'ubuntu-20.04') && (github.event_name == 'pull_request')
+      timeout-minutes: 20
+      run: |
+        cd .github/workflows/docs
+        mkdir extensions
+        ./build-docs -a -i -d ${GITHUB_WORKSPACE}/.github/workflows/docs/extensions
```
cqc-alec, line 83: We should maybe check the output for warnings, as we do for the pytket docs. I see this in the logfile:
```
2022-08-04T09:42:19.4321218Z /opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/pytket/extensions/quantinuum/backends/quantinuum.py:docstring of pytket.extensions.quantinuum.backends.quantinuum.QuantinuumBackend.add_to_batch:3: WARNING: Unexpected indentation.
```

cqc-alec, line 83: The `build-docs` script here could be tidied up a bit: the `-a` option can be removed.

cqc-melf: I have cleaned up the file, will need to check if the upload still works after merging this in a RC release.

cqc-melf: done

cqc-melf: (have removed the i option, too)

cqc-melf:
Had to add an additional file to handle the exit error code in the workflow.

.github/workflows/docs/check-build-docs
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+./.github/workflows/docs/build-docs 1>out.txt 2>err.txt
+cat out.txt err.txt
+ERRS=`cat err.txt | grep -v "start-string without end-string"`
```
cqc-alec, line 5: I don't think this exclusion is necessary when building these docs.

.github/workflows/docs/build-docs
```diff
@@ -80,56 +80,14 @@ def build_module_docs():
...
-    if args.index:
-        print("Building index page...")
-        index_rst = DOCS_DIR / "index.rst"
-        with open(DOCS_DIR / "intro.txt", "r") as f:
-            content = f.readlines()
-            content.append(
-                "\n.. toctree::\n\t:caption: More documentation:\n\t:maxdepth: 1\n\n"
-            )            
-            content.append(f"\tpytket <{PYTKET_DOCS_LINK}>\n")
-            content.append(f"\tpytket extensions <{PYTKET_EX_DOCS_LINK}>\n")
```
cqc-alec, line 109: There is still some clean-up that could be done, e.g. `PYTKET_EX_DOCS_LINK` is no longer needed, though this could be deferred to another PR.

.github/workflows/docs/check-build-docs
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
cqc-alec, line 1: This script is actually failing (but we are swallowing the error from `build-docs`: see the logs.

cqc-alec (changes requested):

.github/workflows/docs/build-docs
cqc-melf: I think this link is still needed? It is still used in line 51, adding the link to the overview of all extensions, see: https://cqcl.github.io/pytket-quantinuum/api/

.github/workflows/docs/check-build-docs
cqc-melf: Will remove it.

.github/workflows/docs/check-build-docs
cqc-melf: Repaired

cqc-melf: Thank you for spotting!

cqc-alec (approved):

PR #23, opened 2022-08-11
[feature] ZZPhase support for devices that support it

lmondada:
As I currently don't have access to any gate fidelity information for the devices, I assume perfect fidelity. This can obviously easily be changed in the future.

Note that I removed the `ZZPhaseToRz` pass when `FullPeephole` is run, as the KAK should already do the optimal thing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ setup(
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket ~= 1.5",
+        "pytket ~= 1.5.2",
         "requests >= 2.2",
         "types-requests",
         "websockets >= 7.0",
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ setup(
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket ~= 1.5.2",
+        "pytket ~= 1.5, >= 1.5.2",
         "requests >= 2.2",
         "types-requests",
         "websockets >= 7.0",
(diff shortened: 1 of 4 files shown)


sjdilkes (dismissed):

lmondada:
Aaargh right, this won't pass until we merge and release https://github.com/CQCL/tket/pull/476. We should probably make a pytket 1.5.1 release for this.

cqc-alec:
I re-ran with pytket 1.5.1, but now some predicate checks are failing in the tests.

cqc-alec:
Pytket 1.5.2 is now released. Could you update the requirement in `setup.py` to `"pytket ~= 1.5, >= 1.5.2"`?

setup.py
```diff
@@ -39,7 +39,7 @@
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket ~= 1.5",
+        "pytket ~= 1.5.2",
```
cqc-alec, line 42: This will restrict to 1.5.x. I think it should be `~= 1.5, >= 1.5.2`, That way 1.6 will be allowed.

cqc-alec: (TBH I'm not completely sure, but I think that's how it works.)

cqc-alec:
I removed `test_zz_phase_support()` because it does not work at optimization level 1 (as I found by hacking the backend object to always support `ZZPhase`). However at optimization level 2 we get the expected result.

cqc-alec:
This now LGTM but let's hold back from merging until we have understood the platform-dependency issue.

cqc-alec:
See https://github.com/CQCL/tket/issues/488 for the underlying issue in TKET. I think this can be merged anyway.

cqc-alec (approved):

PR #28, opened 2022-08-30
Update pytket dependency and tighten test.

cqc-alec:


setup.py
```diff
@@ -39,7 +39,7 @@
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket ~= 1.5, >= 1.5.2",
+        "pytket == 1.6.0rc0",
```
cqc-melf, line 42: Can you add that to the changelog?

cqc-alec: Done.

cqc-melf (approved):

PR #34, opened 2022-09-12
Update link

CalMacCQ:
I have updated the link to the quantinuum web interface to the correct one and also added the quantinuum logo.


.github/workflows/docs/intro.txt
```diff
@@ -1,8 +1,8 @@
 pytket-extensions
 =================
 
-.. image:: CQCLogo.png
-   :width: 120px
+.. image:: Quantinuum_logo.png
+   :width: 200px
```
cqc-melf, line 5: Have you checked if the bigger image is displayed in the right way?

docs/intro.txt
```diff
@@ -1,8 +1,8 @@
 pytket-quantinuum
 ==================================
 
-.. image:: CQCLogo.png
-   :width: 120px
+.. image:: Quantinuum_logo.png
```
cqc-melf, line 4: See above

.github/workflows/docs/intro.txt
```diff
@@ -1,8 +1,8 @@
...
    :align: right
```
cqc-melf, line 8: Do you want to replace the `.. _CQC: https://cambridgequantum.com` link, too?

cqc-melf (changes requested):
Have you added the Quantinuum_logo.png file somewhere? I have not seen that?

.github/workflows/docs/intro.txt
CalMacCQ: Thanks for spotting that - I've replaced it.

.github/workflows/docs/intro.txt
CalMacCQ: Yes I have built the docs locally and it seems like an okay size. Do you think it looks okay?

cqc-melf: I have not taken a look yet, will build it now, too.

.github/workflows/docs/intro.txt
cqc-melf: Perfect.

cqc-melf:
The run of the checks stopped in the middle, I have started a rerun to find out if it works no.

.github/workflows/docs/intro.txt
cqc-melf: Are we allowed within the CI to make the "Q" a bit bigger and keep the "quantinuum" in the same size?

cqc-melf (approved):
From my side this can be merged now. :)

PR #36, opened 2022-09-15
Update example

mlk621:
Updated the basic circuit submission example to include more references and links to pytket documentation. Worked with Kathrin on this. I'll be putting an updated copy on the hardware user portal of that notebook.


ss2165 (changes requested):
* the `available_devices` section lists a load of non-public device names, do we want to keep that? 
* the batching is out of date, see https://cqcl.github.io/pytket-quantinuum/api/index.html#batching
* there should probably be a link to the pytket-quantinuum [docs landing page](https://cqcl.github.io/pytket-quantinuum/api/index.html)  (unless I missed it somewhere). You might also want to think about whether it's worth not repeating some of the information on that page,

mlk621:
@ss2165 

- Ah yes, I forgot to remove that output. I have been given other accesses now I didn't have before, so I didn't have that problem previously. I just deleted the cell output so it isn't shown.
- I had meant to add that link! I added the links to the pytket and pytket-quantinuum docs at the top.
- I updated the batch submission example. Check it is correct. It did work for me.

ss2165 (approved):
Looks good, thanks @mlk621

PR #39, opened 2022-09-15
[feature] Enable custom re-usable sessions for API.

johnchildren:
Adds a session attribute to the API wrapper class which can be supplied when objects of that class are initialized. This is useful when setting custom headers for the session or setting custom retry policies.


pytket/extensions/quantinuum/backends/api_wrappers.py
```diff
@@ -90,6 +90,7 @@ def __init__(
         use_websocket: bool = True,
         provider: Optional[str] = None,
         support_mfa: bool = True,
+        session: Optional[Session] = None,
```
cqc-melf, line 93: Could you please add this to the documentation?

cqc-melf: (And maybe to the changelog in `docs/changelog.rst` )

johnchildren: done!

cqc-melf: Thank you!

cqc-melf (approved):
Looks good, thank you.

PR #41, opened 2022-09-26
Add "options" argument to QuantinuumBackend constructor.

### LINKED ISSUE #33: [feature request] Allow error params to be passed in backnd constructor
These would then be used by process_circuits) by default rather than having them passed in each time.

cqc-alec:
Closes #33 .

### DIFF
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -461,9 +461,9 @@ class QuantinuumBackend(Backend):
                 raise WasmUnsupported("Backend does not support wasm calls.")
             body["cfl"] = wasm_file_handler._wasm_file_encoded.decode("utf-8")
 
+        body["options"].update(self._process_circuits_options)
         if options is not None:
             body["options"].update(options)
-        body["options"].update(self._process_circuits_options)
 
         # apply any overrides or extra options
         body.update(request_options or {})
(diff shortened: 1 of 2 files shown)


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -454,6 +463,7 @@ def submit_qasm(
 
         if options is not None:
             body["options"].update(options)
+        body["options"].update(self._process_circuits_options)
```
cqc-melf, line 466: From my understanding this allows the user to override the previously set options. Is that intended?

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -158,6 +158,7 @@ def __init__(
         provider: Optional[str] = None,
         machine_debug: bool = False,
         api_handler: QuantinuumAPI = DEFAULT_API_HANDLER,
+        **kwargs: QuumKwargTypes,
```
cqc-melf, line 161: Is this a good name for the variable?

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-alec: Yes, but I think I actually have it the wrong way round: the options specified in the backend constructor should be overridden (potentially) by the options passed to `process_circuits()`. I will change this.

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-alec: It's just copied from the `process_circuits()` signature.

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-melf: Perfect, thank you!

cqc-melf (approved):

PR #42, opened 2022-09-28
Update readme

CalMacCQ:
I've updated the README to show the Quantinuum company name rather than CQC.

I've also added a link to the public slack channel and the notebook examples. 

Let me know if anything else should be changed/added/reworded here.


mlk621 (approved):
Looks good!

PR #44, opened 2022-09-29
[test] Add checks for example notebooks

ferbetanzo:
This PR follows the same approach as the `pytket` repo in order to test the example jupyter notebooks in the examples/ folder.

Please note:
- The workflow tests the notebooks with python versions 3.8, 3.9 and 3.10 in parallel, but only in Linux.
- The example notebooks are checked serially.
- We use the `expect` script `answer_input_prompts`, in order to provide the quantinuum credentials interactively, as the user running the notebooks would do. This script has a timeout of 30 minutes since the jobs sent to the emulator can take long to be picked up by the quantinuums' queue (we don't have control over this, so it can happen that the emulator results are not ready before the timeout, in which case the CI would fail).


.pylintrc
```diff
@@ -56,3 +56,5 @@ enable=
 # Ignore long lines containing URLs or pylint or mypy directives.
 ignore-long-lines=^(.*#\w*pylint: disable.*|.*# type: ignore.*|\s*(# )?<?https?://\S+>?)$
 
+# Ignore python scripts that generate example notebooks
+ignore-patterns=Quantinuum_.*.py
```
cqc-alec, line 60: Can we make this match anything in the `examples/python` directory, instead of the filename pattern? Seems more robust that way.

.github/workflows/check-examples.yml
```diff
@@ -0,0 +1,60 @@
...
+    if: github.event_name == 'schedule' || needs.changes.outputs.examples == 'true'
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ['3.8', '3.9', '3.10']
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: '0'
+    - run: git fetch --depth=1 origin +refs/tags/*:refs/tags/* +refs/heads/*:refs/remotes/origin/*
```
cqc-alec, line 45: This is not actually needed, and wastes a bit of time.
```suggestion
```

.github/workflows/check-examples.yml
```diff
@@ -0,0 +1,60 @@
...
+    - cron: '0 4 * * 6'
+
+env:
+  PYTKET_REMOTE_QUANTINUUM_USERNAME: ${{ secrets.PYTKET_REMOTE_QUANTINUUM_USERNAME }}
+  PYTKET_REMOTE_QUANTINUUM_PASSWORD: ${{ secrets.PYTKET_REMOTE_QUANTINUUM_PASSWORD }}
+
+jobs:
+
+  changes:
+    runs-on: ubuntu-20.04
```
cqc-alec, line 19: Can we start using 22.04?

.github/workflows/check-examples.yml
```diff
@@ -0,0 +1,60 @@
...
+    outputs:
+      examples: ${{ steps.filter.outputs.examples }}
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dorny/paths-filter@v2
```
cqc-alec, line 24:
```suggestion
    - uses: dorny/paths-filter@v2.10.2
```
Generally better to specify exact version for external actions when possible.

.github/workflows/check-examples.yml
```diff
@@ -0,0 +1,60 @@
...
+        filters: |
+          examples:
+            - 'examples/**'
+            - '.github/workflows/check-examples.yml'
+
+  check:
+    name: Check notebooks
+    needs: changes
+    if: github.event_name == 'schedule' || needs.changes.outputs.examples == 'true'
+    runs-on: ubuntu-20.04
```
cqc-alec, line 37: 22.04?

.github/workflows/check-examples.yml
```diff
@@ -0,0 +1,60 @@
...
+    strategy:
+      matrix:
+        python-version: ['3.8', '3.9', '3.10']
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: '0'
+    - run: git fetch --depth=1 origin +refs/tags/*:refs/tags/* +refs/heads/*:refs/remotes/origin/*
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
```
cqc-alec, line 47:
```suggestion
      uses: actions/setup-python@v4
```

.github/workflows/check-examples.yml
```diff
@@ -0,0 +1,60 @@
...
+        python-version: ${{ matrix.python-version }}
+    - name: Install required python packages
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install wheel
+        python -m pip install -e .
+        python -m pip install p2j jupyter
+    - name: test example notebooks
+      run: |
+        sudo apt-get install expect
```
cqc-alec, line 58:
```suggestion
      run: |
        sudo apt-get update
        sudo apt-get install expect
```

examples/check-examples
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+for name in `cat maintained-notebooks.txt`
```
cqc-alec, line 5: Do we expect there to be unmaintained notebooks? Ideally not. Can we just test all the notebooks in the directory? Then people can't add one and forget to add it to `maintained-notebooks.txt`.

cqc-alec (changes requested):

.pylintrc
ferbetanzo: I think this is only supported from pylint>=2.9 (https://github.com/PyCQA/pylint/releases/tag/v2.9.0). I'll upgrade our version and see if nothing else breaks :)

.github/workflows/check-examples.yml
ferbetanzo: Will remove :D

.github/workflows/check-examples.yml
ferbetanzo: Sure thing!

examples/check-examples
ferbetanzo: Oh, yes, checking all notebooks is better. I'll take care of it.

ferbetanzo:
@cqc-alec I think I've addressed all the comments :D

cqc-alec (changes requested):

cqc-alec (approved):

PR #46, opened 2022-10-03
update pypi links

cqc-melf:
Update the links on pypi containing bugtracker and documentation: like for example on https://pypi.org/project/wasmer/


cqc-melf:
What do you think about adding this links?

CalMacCQ:
Maybe we should add the links to pypi and github on this side menu as well?

Screenshot 2022-10-03 at 15 14 01

cqc-alec (approved):

PR #47, opened 2022-10-03
add links to bugtracker and github

cqc-melf:
update links, this is was callum suggested in the other PR. I think this is a good idea. Maybe we can find something better then `other` ?


CalMacCQ (approved):

CalMacCQ:
Screenshot 2022-10-03 at 16 40 25

The way the qujax extension looks is not bad.

PR #51, opened 2022-10-06
Fix/changes

cqc-melf:
updates problems found on other repos


.github/workflows/build_and_test.yml
```diff
@@ -46,6 +46,7 @@ jobs:
       with:
         python-version: '3.9'
     - name: Build and test including remote checks (3.9) mypy
+      shell: bash
```
cqc-alec, line 49: Non needed here since this is macos only (but does no harm).

cqc-alec (approved):

PR #57, opened 2022-10-12
update ci workflows to run weekly

cqc-melf:


cqc-melf:
Should be the same PR as for pytket-aqt

cqc-melf:
Is there a reason why the notebook check is not required? (unrelated to this PR) @ferbetanzo

cqc-alec (approved):

PR #72, opened 2022-11-08
[fix] Embed Quantinuum logo in notebooks

ferbetanzo:
The Quantinuum logo is now correctly embedded at the top of the example notebooks. This is done by using the `data:image/svg+xml` in a markdown cell (thx @yao-cqc for the pointer!), and providing the URL encodede `<svg>`. Instead of adding this to each notebook, this embedded logo lives in `logo_header_markdown.md`, and is appended at the top of each notebook in the `check-examples` script.

If you modify the notebook generation scripts (in `examples/python`), make sure to re-generate the corresponding .ipynb by appending the logo at the top, but do not check in the modified python with the header string (otherwise the logo would appear 2 times).


ferbetanzo:
It seems github does not show the embedded svg correctly when previewing the .ipynb file, has this been always the case?
image

examples/check-examples
```diff
@@ -8,10 +8,25 @@ do
...
     rm ${name}-gen.ipynb
-    # Run script:
-    ./answer_input_prompts ${file} $PYTKET_REMOTE_QUANTINUUM_USERNAME $PYTKET_REMOTE_QUANTINUUM_PASSWORD
+    rm python/${file}
+
+    # Restore original
+    mv python/${name}-original.py python/${file}
+
+    # Run script
+    #./answer_input_prompts ${file} $PYTKET_REMOTE_QUANTINUUM_USERNAME $PYTKET_REMOTE_QUANTINUUM_PASSWORD
```
cqc-alec, line 31: Was this meant to be commented out?

cqc-alec (commented):
I think we could do with a `README.md` in the `examples` directory, explaining that the notebooks are genetated from the python files, with any gotchas for contributors.

cqc-alec:
> It seems github does not show the embedded svg correctly when previewing the .ipynb file, has this been always the case?

I'm not sure.

examples/check-examples
ferbetanzo: Good catch! I'll add the README.md :D

yao-cqc:
> It seems github does not show the embedded svg correctly when previewing the .ipynb file, has this been always the case?
> image

It doesn't render in VS code either.

image

yao-cqc:
https://github.com/CQCL/pytket-quantinuum/blob/display-logo-in-notebook/examples/Quantinuum_circuit_submissions.ipynb

I've managed to display the logo by first converting the svg logo to png, and then using the base 64 encode of that png in the markdown cell. Using a base64 encode of the svg doesn't seem to work and I don't know why.

cqc-alec:
> https://github.com/CQCL/pytket-quantinuum/blob/display-logo-in-notebook/examples/Quantinuum_circuit_submissions.ipynb
> 
> I've managed to display the logo by first converting the svg logo to png, and then using the base 64 encode of that png in the markdown cell. Using a base64 encode of the svg doesn't seem to work and I don't know why.

Can we use the PNG instead of SVG?

ferbetanzo:
The png does not look as bad as I thought:
image

ferbetanzo:
@cqc-alec or @yao-cqc, do you think this is ok now?

cqc-alec:
Any idea why the "Check example notebooks" test is failing?

cqc-alec (approved):

yao-cqc (approved):
Looks good. The logo is a little low res, we should ask for a better logo and make an update in a future PR.

PR #73, opened 2022-11-09
Add explicit test for conditional phase

cqc-alec:
Also bump pytket requirement to 1.9.0rc0.

### DIFF
--- a/.github/workflows/check-examples.yml
+++ b/.github/workflows/check-examples.yml
@@ -28,6 +28,7 @@ jobs:
         filters: |
           examples:
             - 'setup.py'
+            - 'pytket/**'
             - 'examples/**'
             - '.github/workflows/check-examples.yml'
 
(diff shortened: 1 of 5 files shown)


cqc-melf (approved):

.github/workflows/check-examples.yml
```diff
@@ -27,6 +27,7 @@ jobs:
         base: ${{ github.ref }}
         filters: |
           examples:
+            - 'setup.py'
```
cqc-melf, line 30: We probably should check for any changes that are not only updating the docs or the the tests? (So include here pytket/* )?

cqc-melf, line 30:
```suggestion
            - 'setup.py'
            - 'pytket/**'
```

cqc-melf (approved):

PR #74, opened 2022-11-15
Use `actions/deploy-pages` to publish docs

cqc-alec:
This is a new way to publish to github pages that doesn't use the special `gh-pages` branch.

It is still officially "beta" but seems to work.

Requires a modification to the repo "Pages" settings to enable it. (I modified the settings to test it, and have now reverted them, but will switch back once this is merged).

Tested here: https://github.com/CQCL/pytket-quantinuum/actions/runs/3468800329
Docs successfully uploaded: https://cqcl.github.io/pytket-quantinuum/api/index.html

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -109,8 +109,8 @@ jobs:
         password: ${{ secrets.PYPI_PYTKET_QUANTINUUM_API_TOKEN }}
         verbose: true
 
-  docs:
-    name: Build and publish docs
+  build_docs:
+    name: Build docs
     if: github.event_name == 'release'
     needs: publish_to_pypi
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


cqc-melf (commented):
Are there reasons why this is still beta documented somewhere?

Is there still a way edit the docs directly? Can this now be done on the main branch in that directory?

cqc-alec:
> Are there reasons why this is still beta documented somewhere?

Not that I've seen, I assume it is just that it is fairly new and they wanted to make sure it was all fine before moving out of beta. They must be pretty confident though because it was made the default method in August: https://github.blog/changelog/2022-08-10-github-pages-builds-with-github-actions-ga/ .

> Is there still a way edit the docs directly? Can this now be done on the main branch in that directory?

No it is not possible to edit them on a branch and push that branch: the docs (or anything deployed using this method) are no longer stored in source control; they are generated on the CI. So it is purely for autogenerated content. In principle if you wanted to make an edit you could do so by scripting that edit into the job that generates the docs, but this is not very nice.

cqc-melf:
What I was manly thinking about is the `.nojekyll` which we need for some of the extensions in the doc build, Is that something you needed to do in the setting or is that not required anymore?

cqc-melf (approved):
Great! I am happy with this now, and I think we should add this in the same way to the other repos. What do you mean?

cqc-alec:
> Great! I am happy with this now, and I think we should add this in the same way to the other repos. What do you mean?

Yeah that was my hope!

PR #75, opened 2022-11-17
add pytket upgrade to build file

cqc-melf:
add upgrade to newest pytket version (including pre release for every run) and run on every push a branch named "runci/..."

### DIFF
--- a/.github/workflows/build-test
+++ b/.github/workflows/build-test
@@ -51,6 +51,7 @@ cd ${GITHUB_WORKSPACE}/tests
 
 python -m pip install --pre -r test-requirements.txt
 
+# update the pytket version to the lastest (pre) release
 python -m pip install --upgrade --pre pytket~=1.0
 
 pytest --doctest-modules
--- a/.github/workflows/build-test
+++ b/.github/workflows/build-test
@@ -51,6 +51,8 @@ cd ${GITHUB_WORKSPACE}/tests
 
 python -m pip install --pre -r test-requirements.txt
 
+python -m pip install --upgrade --pre pytket
+
 pytest --doctest-modules
 
 cd ..
--- a/.github/workflows/build-test
+++ b/.github/workflows/build-test
@@ -51,7 +51,7 @@ cd ${GITHUB_WORKSPACE}/tests
 
 python -m pip install --pre -r test-requirements.txt
 
-python -m pip install --upgrade --pre pytket
+python -m pip install --upgrade --pre pytket~=1.0
 
 pytest --doctest-modules
 
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -45,7 +45,6 @@ jobs:
       run: |
         ./.github/workflows/build-test nomypy
     - name: Set up Python 3.9
-      if: github.event_name == 'push' || github.event_name == 'pull_request' || github.event_name == 'release' || github.event_name == 'schedule'
       uses: actions/setup-python@v4
       with:
         python-version: '3.9'
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -45,19 +45,19 @@ jobs:
       run: |
         ./.github/workflows/build-test nomypy
     - name: Set up Python 3.9
-      if: github.event_name == 'push' || github.event_name == 'pull_request' || github.event_name == 'release' || contains(github.ref, 'refs/heads/wheel') || github.event_name == 'schedule'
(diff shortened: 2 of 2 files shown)


cqc-melf:
If you are happy with this I will add the same changes to all other extension repos

.github/workflows/build_and_test.yml
```diff
@@ -44,19 +45,19 @@ jobs:
       run: |
         ./.github/workflows/build-test nomypy
     - name: Set up Python 3.9
-      if: github.event_name == 'pull_request' || github.event_name == 'release' || contains(github.ref, 'refs/heads/wheel') || github.event_name == 'schedule'
+      if: github.event_name == 'push' || github.event_name == 'pull_request' || github.event_name == 'release' || contains(github.ref, 'refs/heads/wheel') || github.event_name == 'schedule'
```
cqc-alec, line 48:
```suggestion
      if: github.event_name == 'push' || github.event_name == 'pull_request' || github.event_name == 'release' || github.event_name == 'schedule'
```
This is redundant I think (also below).

.github/workflows/build-test
```diff
@@ -51,6 +51,8 @@ cd ${GITHUB_WORKSPACE}/tests
 
 python -m pip install --pre -r test-requirements.txt
 
+python -m pip install --upgrade --pre pytket
```
cqc-alec, line 54: We should probably restrict to pytket 1.x.

cqc-alec (changes requested):
I

.github/workflows/build-test
cqc-melf: Good point

.github/workflows/build_and_test.yml
cqc-melf: Yes, I have removed that.

cqc-melf:
If you are happy I will start with the other PRs before the remote checks are finished at this one?

.github/workflows/build_and_test.yml
```diff
@@ -44,19 +45,19 @@ jobs:
       run: |
         ./.github/workflows/build-test nomypy
     - name: Set up Python 3.9
-      if: github.event_name == 'pull_request' || github.event_name == 'release' || contains(github.ref, 'refs/heads/wheel') || github.event_name == 'schedule'
+      if: github.event_name == 'push' || github.event_name == 'pull_request' || github.event_name == 'release' || github.event_name == 'schedule'
```
cqc-alec, line 48: I guess this line can be removed entirely now since it will always be true.

.github/workflows/build-test
```diff
@@ -51,6 +51,8 @@ cd ${GITHUB_WORKSPACE}/tests
 
 python -m pip install --pre -r test-requirements.txt
 
+python -m pip install --upgrade --pre pytket~=1.0
```
cqc-alec, line 54: Perhaps a comment here would be good, just to explain that we want to test against pre-release versions.

cqc-alec (changes requested):

.github/workflows/build_and_test.yml
cqc-melf: done

.github/workflows/build-test
cqc-melf: done

.github/workflows/build_and_test.yml
```diff
@@ -44,19 +45,18 @@ jobs:
...
+      if:  (matrix.os == 'macos-12') && (github.event_name == 'push' || (github.event_name == 'pull_request' &&  github.event.pull_request.head.repo.full_name == github.repository) || github.event_name == 'release' || github.event_name == 'schedule' )
       run: |
         ./.github/workflows/build-test mypy
       env:
         PYTKET_RUN_REMOTE_TESTS: 1
     - name: Build and test including remote checks (3.9) nomypy
-      if:  (matrix.os != 'macos-12') && ((github.event_name == 'pull_request' &&  github.event.pull_request.head.repo.full_name == github.repository) || github.event_name == 'release' || contains(github.ref, 'refs/heads/wheel') || github.event_name == 'schedule')
+      if:  (matrix.os != 'macos-12') && (github.event_name == 'push' || (github.event_name == 'pull_request' &&  github.event.pull_request.head.repo.full_name == github.repository) || github.event_name == 'release' || github.event_name == 'schedule')
```
cqc-melf, line 59: @cqc-alec Maybe we can discuss this here in this PR: I am overwriting the condition for external PRs on purpose because that is just stopping the tests to run and it would be better if we just take over external PRs to make sure that all checks are run. (like I did for AQT)
Is that okay or should we add this condition for all repos again?

PR #81, opened 2022-11-23
update example in preparation of TKET compilation integration to H-series stack

mlk621-qtm:


mlk621-qtm:
@ferbetanzo : I saw the update to pytket to have default compilation pass of optimization level 2 in get_compiled_circuits, so updated description to reflect this as well as figured out running black for the code formatting. Its still failing, but looks like something related to the other notebooks. I only touched Quantinuum_circuit_submission.ipynb via running p2j, as instructions didn't mention the role of the other notebooks in the README in the examples folder so if there's something else to do please let me know.

examples/python/Quantinuum_circuit_submissions.py
```diff
@@ -167,6 +167,7 @@
 
 # Jobs that have been submitted can also be cancelled if needed.
 
+handle = backend.process_circuit(compiled_circuit, n_shots=n_shots)
```
yao-cqc, line 170: Otherwise the next line will raise an error saying the job has already been completed.

mlk621-qtm: ok, I commented out that line, this is where using this automated thing on the script instead of the notebook itself gets a little goofy. there are places in the original notebooks that have comments in the code cells, but they get converted to markdown because of `p2j` detecting them as such since the lines starts with a #. And in this case, the best thing is to show the user what the function is in a code cell. I can see how it causes issue here, ... but that's just my thought on using a script to check that the notebooks are up to date... not sure how to get around that

yao-cqc (approved):

PR #82, opened 2022-11-24
Add CustomPass for flattening registers to default compilation

sjdilkes:


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -411,6 +423,7 @@ def default_compilation_pass(self, optimisation_level: int = 2) -> BasePass:
                     SimplifyInitial(
                         allow_classical=False, create_all_qubits=True, xcirc=_xcirc
                     ),
+                    CustomPass(flatten_registers),
```
cqc-alec, line 426: Since this is done at all optimization levels, can we just add it unconditionally that the end instead of inside each branch?

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -366,6 +366,11 @@ def default_compilation_pass(self, optimisation_level: int = 2) -> BasePass:
         ]
         squash = auto_squash_pass({OpType.PhasedX, OpType.Rz})
 
+        def flatten_registers(c: "Circuit") -> "Circuit":
```
cqc-alec, line 369: It would be good to add a comment explaining why we are doing this. Particularly why it is necessary here but not in other backends (or is it?).

tests/backend_test.py
```diff
@@ -222,12 +222,18 @@ def test_default_pass(
...
+        c.add_qubit(q0)
+        c.H(q0)
         c.H(0)
         c.CX(0, 1)
         c.CSWAP(1, 0, 2)
         c.ZZPhase(0.84, 2, 0)
         c.measure_all()
+        c.add_qubit(q1)
         comp_pass.apply(c)
+        assert c.qubits == [Qubit(0), Qubit(1), Qubit(2), Qubit(3)]
```
cqc-alec, line 236: Is it possible to add a test highlighting the issue this is fixing?

cqc-alec (changes requested):

pytket/extensions/quantinuum/backends/quantinuum.py
sjdilkes: I can do if you think the QASM explanation is lacking though!

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -366,6 +366,24 @@ def default_compilation_pass(self, optimisation_level: int = 2) -> BasePass:
...
+        # Both of these cases can causes issues when converting to QASM,
+        # as the size of the defined "qreg" can be larger than the
+        # number of Qubits actually used, or at times larger than the
+        # number of device Qubits, even if fewer are really used.
+        # By flattening the Circuit qubit registers, we make sure
+        # that the produced QASM has one "qreg", with the exact number
+        # of qubits actually used in the Circuit.
+        def flatten_registers(c: "Circuit") -> "Circuit":
+            c.remove_blank_wires()
+            c.rename_units({c.qubits[i]: Qubit(i) for i in range(len(c.qubits))})
```
yao-cqc, line 384: My only concern is that the renaming map is not visible to the user. Since we are not returning the map to the user maybe we should add a note explain how the renaming is done (i.e. c.qubits[i] -> Qubit(i))?

sjdilkes: Happy to add a note explaining this. The reason I haven't is that we also don't provide that information when we map to a physical Circuit in other devices. I'll also change from `Qubit(i)` to `Qubit("quantinuum", i)`  to make this clearer.

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-alec: Looks like it is still being added in the individual branches?

sjdilkes: Oops, done.

pytket/extensions/quantinuum/backends/quantinuum.py
yao-cqc: Thanks, I think that should be sufficient. 

In other devices, shouldn't the user still be able to know the placement by wrapping the circuit in a `CompilationUnit` and compile the unit with the `get_default_compilation_pass` ? I might be completely wrong.

sjdilkes: Ah ok, yes they can do that. So the concern is that `CustomPass` might not update the `maps` attribute suitably? Maybe we can raise a JIRA ticket to add a flattening pass to TKET for a followup?

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -366,6 +366,28 @@ def default_compilation_pass(self, optimisation_level: int = 2) -> BasePass:
...
+        # Both of these cases can causes issues when converting to QASM,
+        # as the size of the defined "qreg" can be larger than the
+        # number of Qubits actually used, or at times larger than the
+        # number of device Qubits, even if fewer are really used.
+        # By flattening the Circuit qubit registers, we make sure
+        # that the produced QASM has one "qreg", with the exact number
+        # of qubits actually used in the Circuit.
+        # The Circuit qubits attribute is iterated through, with the ith
+        # Qubit being assigned to the ith qubit of a new "quantinuum" register
+        def flatten_registers(c: "Circuit") -> "Circuit":
```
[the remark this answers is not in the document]
sjdilkes: done

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -104,6 +104,12 @@ def _get_gateset(gates: List[str]) -> Set[OpType]:
     return gs
 
 
+def _flatten_registers(c: "Circuit") -> "Circuit":
+    c.remove_blank_wires()
+    c.rename_units({c.qubits[i]: Qubit("quantinuum", i) for i in range(len(c.qubits))})
```
cqc-alec, line 109:
```suggestion
    c.rename_units({qb: Qubit("quantinuum", i) for i, qb in enumerate(c.qubits)})
```
Just a suggestion for style and efficiency, feel free to ignore.

sjdilkes: added

cqc-alec:
We should add this change to the changelog, so that people are not (too) surprised by it.

pytket/extensions/quantinuum/backends/quantinuum.py
sjdilkes: done

pytket/extensions/quantinuum/backends/quantinuum.py
yao-cqc: Sounds good!

PR #83, opened 2022-11-25
update doc build to use github deployment

cqc-melf:


cqc-melf:
Would like to update this to make sure that the workflows are as similar as possible

yao-cqc (approved):

cqc-alec (approved):

PR #94, opened 2022-12-20
Ensure tests are ready for H2

cqc-alec:
When the H2 is released we should add more tests for it; this PR is just to make sure the existing tests continue to work. (Also add H2-1 to the "offline" API, and remove an unused variable.)

### DIFF
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -66,7 +66,6 @@ from pytket.extensions.quantinuum.backends.credential_storage import (
 from .api_wrappers import QuantinuumAPIError, QuantinuumAPI
 
 _DEBUG_HANDLE_PREFIX = "_MACHINE_DEBUG_"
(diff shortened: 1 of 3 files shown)


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -66,8 +66,6 @@
 from .api_wrappers import QuantinuumAPIError, QuantinuumAPI
 
 _DEBUG_HANDLE_PREFIX = "_MACHINE_DEBUG_"
-QUANTINUUM_URL_PREFIX = "https://qapi.quantinuum.com/"
```
cqc-melf, line 69: What happend to the url?

cqc-alec: This is defined here: https://github.com/CQCL/pytket-quantinuum/blob/develop/pytket/extensions/quantinuum/backends/api_wrappers.py#L77

The variable `QUANTINUUM_URL_PREFIX` is never used.

cqc-melf (approved):

PR #99, opened 2023-01-04
Add script to generate notebooks

yao-cqc:
This should make script generation much easier.

### DIFF
--- a/examples/check-examples
+++ b/examples/check-examples
@@ -28,5 +28,5 @@ do
     mv python/${name}-original.py python/${file}
 
     # Run script
-    # ./answer_input_prompts ${file} $PYTKET_REMOTE_QUANTINUUM_USERNAME $PYTKET_REMOTE_QUANTINUUM_PASSWORD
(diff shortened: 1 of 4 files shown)


examples/check-examples
```diff
@@ -2,12 +2,13 @@
 
 set -e
 
-for file in $(ls python/*py | xargs -n 1 basename)
+for name in `cat maintained-notebooks.txt`
```
ferbetanzo, line 5: Do you think we would have notebooks that are generated with python but that we'd like to skip in the `check-examples` and `gen-nb` scripts? If not, I think getting all the scripts under python/ is better.

yao-cqc: Yeah, I think we should check all notebooks at least for now. Do you think we should remove ``maintained-notebooks.txt`` all together? If we don't keep unchecked notebooks then I'm not sure what is the point of this file. We can always bring this file back when needed.

ferbetanzo: Yeah, I think removing `maintained-notebooks.txt` is good :)

ferbetanzo (approved):
Looks great, thx!

PR #100, opened 2023-01-05
update expect script to only send inputs if needed

yao-cqc:
answer_input_prompts would fail if a python script doesn't prompt  for inputs.


ferbetanzo (approved):
looks good :)

PR #127, opened 2023-04-11
add kwargs to cost calculation

cqc-melf:
add kwargs to cost calculation to hand them over to the process circuit function


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -875,7 +879,7 @@ def cost(
             cast(str, syntax_checker), api_handler=self.api_handler
         )
         try:
-            handle = backend.process_circuit(circuit, n_shots)
+            handle = backend.process_circuit(circuit, n_shots, kwargs=kwargs)  # type: ignore
```
cqc-melf, line 882: ignore required because mypy is comparing this to the signature of the pytket backend kwargs types.

yao-cqc (approved):
Looks good!

PR #131, opened 2023-04-27
Main

cqc-melf:


cqc-alec (commented):
I guess we shouldn't merge this, since we didn't do the release.

cqc-melf:
The check only failed because of an issues with the backend. Should be fine after rerunning.
I think we should merge this PR or revert the commits on main to avoid conflicts with the next release.

cqc-alec:
> The check only failed because of an issues with the backend. Should be fine after rerunning. I think we should merge this PR or revert the commits on main to avoid conflicts with the next release.

I thought the reason we weren't releasing was because `MaxNClRegPredicate` isn't in pytket 1.14? If that is the case I don't get how the tests are passing?

(I don't think there should be any merge conflicts if we keep the same release branch for the next merge.)

cqc-melf:
The ci is always using the latest pytket release which picks the 1.15. RC, which includes the missing predicate.

The question is if we want to have a dependency on an pytket rc version? Which is something we should not do from my understanding.

cqc-melf:
The tests are all fine now

cqc-melf:
If we are doing new changes to for example the changelog on develop we will have conflicts on the release branch as well when we are updating it with the new changes. I think we should merge this or do a revert on main.

cqc-alec:
> The ci is always using the latest pytket release which picks the 1.15. RC, which includes the missing predicate.

Oh yes I forgot that the CI picks up the latest pre-release!

> The question is if we want to have a dependency on an pytket rc version? Which is something we should not do from my understanding.

No we shouldn't -- so we should hold back on the release.

cqc-alec (approved):

cqc-melf:
Sounds good, I will do that

PR #135, opened 2023-05-04
Add get_partial_result

yao-cqc:
Websocket only message back on status change, therefore it is not useful for retrieving partial results. I decided to add a simple method that makes a single call to the endpoint. It's up to the user to decide how they want to use this method and process data.


yao-cqc:
It seems H1-2E isn't responding. Will rerun the tests

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -779,6 +779,32 @@ def circuit_status(
...
+        self, handle: ResultHandle
+    ) -> Tuple[Optional[BackendResult], CircuitStatus]:
+        """
+        Retrieve partial results for a given job, regardless of its current state.
+
+        :param handle: handle to results
+        :type handle: ResultHandle
+
+        :return: A tuple containing the results and circuit status.
+        If no results are available, the first element is None.
```
cqc-alec, line 792: Do we need an indent here?

cqc-alec (approved):
Looks good!

cqc-melf (approved):

PR #139, opened 2023-05-10
Feature/support rxxyyzz gate

yao-cqc:


cqc-melf:
Do we want this in the release?

yao-cqc:
> Do we want this in the release?

Yes, if we are not in a hurry and this PR doesn't require too much change.

cqc-melf:
Yes we can do that. I would suggest that we merge this directly into the release branch as well?

cqc-alec (approved):

PR #141, opened 2023-05-18
Experimental QIR support

cqc-alec:
For "experimental" read "not working". However, at least `QuantinuumBackend.submit_program()` with valid QIR bitcode _is_ working, and the plumbing is there for `process_circuits()` (though if you try to use it now you'll get a warning and the submission will almost certainly fail).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ setup(
     include_package_data=True,
     install_requires=[
         "pytket ~= 1.15",
-        "pytket-qir ~= 0.1.5",
+        "pytket-qir == 0.2.0rc1",
         "requests >= 2.2",
         "types-requests",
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -44,6 +44,7 @@
     include_package_data=True,
     install_requires=[
         "pytket ~= 1.15",
+        "pytket-qir == 0.2.0rc1",
```
cqc-melf, line 47: Do we want to have this as a dependency or do we want to do something like ask the user to install it if they want to use this feature? I would assume that this could generated problems, because pytket-qir is not supporting arm64 linux?

cqc-melf (commented):
Looks good, one short question.

cqc-melf (approved):

PR #150, opened 2023-06-06
explain partial results

CalMacCQ:
Added a note to the docs showing how to make use of partial results.

I've also tried to give some motivating use cases.


yao-cqc (approved):
LGTM

PR #151, opened 2023-06-06
Docs/minor cleanup

CalMacCQ:
minor edits to documentation

updating copright and author name in documentation footer
correct tket support email
deleted an unused file

Updated sphinx requirement. We can use any version up to but not including 6.2.0.


yao-cqc (approved):
Thanks!

PR #155, opened 2023-06-09
Truncate results from QIR processing to desired width

cqc-alec:
Add an extra field to the `ResultHandle` to track this. While here, make extraction of values from result handles a little less error-prone by adding and using dedicated methods.

### DIFF
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -448,9 +448,10 @@ class QuantinuumBackend(Backend):
 
     @property
     def _result_id_type(self) -> _ResultIdTuple:
-        return tuple((str, str))
+        return tuple((str, str, int))
 
-    def get_jobid(self, handle: ResultHandle) -> str:
+    @staticmethod
+    def get_jobid(handle: ResultHandle) -> str:
         """Return the corresponding Quantinuum Job ID from a ResultHandle.
 
         :param handle: result handle.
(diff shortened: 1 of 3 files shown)


tests/backend_test.py
```diff
@@ -928,4 +928,6 @@ def test_qir_conversion(authenticated_quum_backend: QuantinuumBackend) -> None:
     c = b.get_compiled_circuit(c0)
     h = b.process_circuit(c, n_shots=10, language=Language.QIR)  # type: ignore
     r = b.get_result(h)
-    assert len(r.get_shots()) == 10
+    shots = r.get_shots()
+    assert len(shots) == 10
+    assert all(len(shot) == 2 for shot in shots)
```
cqc-melf, line 933: What would be the expected result for a circuit containing more than one classical register? Do we want to add tests for that?

cqc-alec: I'd expect it to work, but we cannot test it at the moment becaiuse of https://github.com/CQCL/pytket-qir/blob/f854760ec4d9ed868bfd6de2d8978f1718ba082c/pytket/qir/conversion/api.py#L88 .

cqc-melf: Good point, we should keep that in mind and add a test after this compatibility issues are resolved.

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -460,6 +461,30 @@ def get_jobid(self, handle: ResultHandle) -> str:
...
+        :param handle: result handle
+        :return: serialized post-processing circuit, if any
+        """
+        return json.loads(cast(str, handle[1]))
+
+    @staticmethod
+    def get_results_width(handle: ResultHandle) -> Optional[int]:
+        """Return the truncation width of the results, if any.
+
+        :param handle: result handle
```
cqc-melf, line 478: documentation of the type is missing

cqc-melf, line 478:
```suggestion
        :param handle: result handle
        :type handle: ResultHandle

File: pytket/extensions/quantinuum/backends/quantinuum.py [hunk 3]
```diff
@@ -460,6 +461,30 @@ def get_jobid(self, handle: ResultHandle) -> str:
         """
         return cast(str, handle[0])
 
+    @staticmethod
+    def get_ppcirc_rep(handle: ResultHandle) -> Any:
+        """Return the JSON serialization of the classiocal postprocessing circuit
+        attached to a handle, if any.
+
+        :param handle: result handle
```
cqc-melf, line 469:
```suggestion
        :param handle: result handle
        :type handle: ResultHandle

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-alec: I prefer not to add it to the docstring as it's already in the function signature (which is visible in the published docs), so it would just be duplicating information.

cqc-melf (approved):

PR #156, opened 2023-06-12
[test] Separate integration and unit tests

### LINKED ISSUE #147: Separate unit and integration tests
Right now the tests in `/tests` include both unit and integration tests. Some of the latter ones take too long to complete as they target the quantinuum simulators. We should separate them and decide when and how often to run the integration tests in the CI.

ferbetanzo:
Fixes #147.

The integration (remote) tests are only run with python 3.10 in the following events:
- push to main or develop.
- PR (internal).
- scheduled.
- manually dispatched (specifying 'yes' in the input).

The examples checks are run when:
- scheduled.
- push to main.
- PR to main or develop (if any file in `examples/**` was modified).


yao-cqc (commented):
Should the old files in `tests` be removed?

.github/workflows/build_and_test.yml
```diff
@@ -48,27 +56,103 @@ jobs:
...
+        matrix.os == 'macos-12' &&
+        (
+          (
+            github.event_name == 'push' &&
+            (
+              contains(github.ref_name, 'main') ||
+              contains(github.ref_name, 'runci/')
+            )
+          ) ||
+          github.event_name == 'schedule' ||
```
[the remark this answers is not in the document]
ferbetanzo: Sorry, did not notice that the files were not moved but created as new. I've removed the old ones.

Anmd for the `schedule` event, good catch. I've fixed it too. :)

tests/integration/backend_test.py
cqc-alec, line 1: Not all of the tests in `backend_test.py` require remote access; these could be included in the `unit` directory.

cqc-alec, line 1: Two of the tests in `backen_test.py` are annotated with

```python
@pytest.mark.skip("Will be moved to a set of long running integration tests")
```

These are still never run; should they be run on a schedule?

ferbetanzo: I've created a new `unit/offline_backend_test.py` which contains the backend tests that do not require remote access.

ferbetanzo: Indeed, good catch! I've enabled these 2 tests.

cqc-alec (approved):

PR #164, opened 2023-06-21
Release/0.17.1rc0

cqc-melf:


cqc-alec:
Don't we normally do the rc releases from `develop`?

cqc-melf:
> Don't we normally do the rc releases from `develop`?

Normally yes, but I am doing this one to find a problem in the workflow that is not showing up when running it from develop.

cqc-alec:
> 

What is the problem?

PR #167, opened 2023-06-22
Run some existing tests with QIR as well as QASM submission

cqc-alec:
In some places I've commented out the QIR option and added a FIXME comment.

### DIFF
--- a/tests/integration/backend_test.py
+++ b/tests/integration/backend_test.py
@@ -159,12 +159,13 @@ def test_bell(
     "language",
     [
         Language.QASM,
-        # Language.QIR, # FIXME enable when multiregister circuits supported
+        Language.QIR,
     ],
 )
 def test_multireg(
     authenticated_quum_backend: QuantinuumBackend, language: Language
 ) -> None:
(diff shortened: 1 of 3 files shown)


cqc-melf:
We maybe want to try to set this to false for the test as well?
https://github.com/CQCL/pytket-quantinuum/blob/develop/pytket/extensions/quantinuum/backends/quantinuum.py#LL688C33-L688C37

cqc-alec:
> We maybe want to try to set this to false for the test as well? https://github.com/CQCL/pytket-quantinuum/blob/develop/pytket/extensions/quantinuum/backends/quantinuum.py#LL688C33-L688C37

Good point. But `test_qir_conversion` will fail if we set it to `False`.

Maybe we should wait for Pete's upgrade, them remove that option, before merging this. I really just wanted to push this branch so as to be able to refer to specific tests when raising issues.

cqc-melf:
This PR does not run all the checks, reopend at: https://github.com/CQCL/pytket-quantinuum/pull/175

Issue #181, opened 2023-07-11
Allow implicit swaps in default rebase pass


cqc-alec:
After https://github.com/CQCL/tket/issues/931 .

PR #182, opened 2023-07-14
Update optimisation level 0 to add implicit swap gates

sjdilkes:
This PR is one way of satisfying https://github.com/CQCL/pytket-quantinuum/issues/181.

Some operations, such as `OpType.SWAP`, `OpType.ISWAPMax` and `OpType.Sycamore` have the effect of performing a wire swap in a Circuit. Given this they can be decompose in different ways, either correcting for the wire swap with operations or allowing it to be performed "implicitly" by rewriting the DAG.

We would like optimisation level 0 to allow implicit swaps when compiling circuits, but not perform any further optimisation.

Optimisation levels 1 and 2 allow implicit wire swaps already: they resynthesise all two-qubit gates into TK2, merge them and then Decompose them into the allowed primitives (here ZZMax) with `allow_swaps = True`. In this way they produced optimised circuits with implicit wire swaps.

Optimisation level 0 does not do this, only performing a rebase of the Circuit into {Rz, PhasedX, ZZMax}.

Note that a rebase pass works as follows:
1) Convert all gates to TK1 or TK2 without optimisation
2) Convert TK1 and TK2 gates to desired gate set via pre-defined Circuit definitions

Therefore we can emulate rebasing with implicit wire swaps without optimisation by the given compiler passes:
1) rebase the Circuit to {Rz, PhasedX, TK2} (does not merge or optimise TK2 gates)
2) NormaliseTK2() (a Compiler Pass precondition, should not effect TK2 produced from ISWAPMax, Sycamore or SWAP given how they are converted to TK2)
3) DecomposeTK2(), with `allow_swaps` = True, adding implicit swap rewiring
4) The usual rebase_pass, compiling to {Rz, PhasedX, ZZMax}

### DIFF
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -399,7 +399,14 @@ class QuantinuumBackend(Backend):
         # https://cqcl.github.io/pytket-quantinuum/api/index.html#default-compilation
         # Edit this docs source file -> pytket-quantinuum/docs/intro.txt
         if optimisation_level == 0:
-            passlist.append(self.rebase_pass())
+            passlist.extend(
+                [
+                auto_rebase_pass({OpType.Rz, OpType.PhasedX, OpType.TK2}),
+                NormaliseTK2(),
+                DecomposeTK2(),
+                self.rebase_pass(),
+                ]
+            )
         elif optimisation_level == 1:
             passlist.extend(
                 [
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -408,7 +408,14 @@ class QuantinuumBackend(Backend):
             passlist.extend(
                 [
                     CustomPass(lambda c: replace_swaps(c)),
-                    auto_rebase_pass({OpType.Rz, OpType.PhasedX, OpType.TK2}),
+                    auto_rebase_pass(
+                        {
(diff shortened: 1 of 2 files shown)


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -398,8 +398,22 @@ def default_compilation_pass(self, optimisation_level: int = 2) -> BasePass:
...
-            passlist.append(self.rebase_pass())
+
+            def replace_swaps(c: Circuit) -> Circuit:
+                c.replace_SWAPs()
+                return c
+
+            passlist.extend(
+                [
+                    CustomPass(lambda c: replace_swaps(c)),
+                    auto_rebase_pass({OpType.Rz, OpType.PhasedX, OpType.TK2}),
```
yao-cqc, line 411: This will break native gates, for example
```python
b = QuantinuumBackend("", machine_debug=True)
c = Circuit(2).ZZMax(0, 1)
d = b.get_compiled_circuit(c, optimisation_level=0)
d.n_gates #26
```

yao-cqc (approved):
Thanks!

sjdilkes:
> Thanks!

I'm going to hold off merging this as I've found another issue, which is that the number of single qubit wrapper gates are excesssive, due to being decompose twice and not merged. I'm not currently sure how to fix this at the level of compiler passes: I may need to instead go to `TKET` core and update `auto_rebase_pass` to allow implicit SWAP gates instead.

PR #187, opened 2023-08-02
Nb updates

mlk621-qtm:
Updated python notebooks to align with the user portal versions, updated Wasm notebook to make it little clearer/cleaner and match formatting of other notebooks


mlk621-qtm:
@yao-cqc I made some updates to the notebooks to align with what is on the hardware user portal. I think the `gen-nb` script / process for creating the notebooks changed since I last contributed. I'm not sure the script is going to run for me since I'm on Windows? Can you ensure to run the notebook creation script and commit that before accepting this? thanks!

yao-cqc (approved):

PR #193, opened 2023-08-09
Forward compatibility of `ResultHandle`

### LINKED ISSUE #189: Ensure `ResultHandle` from old version of extension can be used with new version
Recently we changed the internal format of `ResultHandle` and this meant old saved ones weren't usable with new versions of pytket-quantinuum.

To avoid this problem in future, we should either promise backward-compatibility or offer an explicit upgrade path.

cqc-alec:
Closes #189 .


yao-cqc (approved):
Looks good!

Issue #192, opened 2023-08-09
Leakage-detection test failing


cqc-alec:
The leakage-detection tests are not being run on the CI, because the file `leakage_detection_test.py` is not in the `unit/` directory.

One of the tests (`test_postselection_circuits_2qb_1_spare_task_gen`) is failing locally.

cqc-alec:
Fixed by #194 .

PR #194, opened 2023-08-09
Always measure leakage detection result to a new Bit

sjdilkes:
To fix -> https://github.com/CQCL/pytket-quantinuum/issues/192
We use spare device qubits for leakage detection. We also reuse these spare device qubits for multiple leakage detections if needed (though reusing each qubit as few times as possible). I accidentally repeated the logic for "resetting" this spare device qubit for also assigning Bit to send measurement results to - this PR makes sure all leakage detection have their own Bit.


yao-cqc (approved):
Looks good!

PR #205, opened 2023-08-17
update pytket-qir to 0.2.0

cqc-melf:
update to pytket-qir 0.2.0, this should only be merged after the result handling issue is solved.


.github/workflows/build-test
```diff
@@ -63,7 +63,7 @@ python -m pip install --pre -r test-requirements.txt
 python -m pip install --upgrade --pre pytket~=1.0
 
 # install the required pytket-qir version
-python -m pip install pytket-qir==0.2.0rc14
+python -m pip install --upgrade --pre pytket~=0.2.0
```
cqc-melf, line 66:
```suggestion
python -m pip install --upgrade --pre pytket-qir~=0.2.0

File: tests/integration/backend_test.py [hunk 2]
```diff
@@ -374,7 +394,7 @@ def test_classical(
     b = authenticated_quum_backend
 
     c = b.get_compiled_circuit(c)
-    assert b.run_circuit(c, n_shots=10).get_counts()
+    assert b.run_circuit(c, n_shots=10, language=language).get_counts()  # type: ignore
```
cqc-alec, line 397: Not sure why we need these `type:ignore` now; is `language` not properly typed?

tests/integration/backend_test.py
```diff
@@ -652,8 +681,15 @@ def test_device_state(
 @pytest.mark.parametrize(
     "authenticated_quum_backend", [{"device_name": "H1-1SC"}], indirect=True
 )
+@pytest.mark.parametrize(
+    "language",
+    [
+        Language.QASM,
+        # Language.QIR, # FIXME needs WASM support
```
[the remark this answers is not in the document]
cqc-alec: Is there a github issue we can reference here?

cqc-alec (changes requested):

tests/integration/backend_test.py
cqc-melf: It is because all the parameters here are not individually defined, they are handled as dict:
```tests/integration/backend_test.py:405: error: Argument "language" to "run_circuit" of "Backend" has incompatible type "Language"; expected "Union[int, float, str, None]"  [arg-type]```

run_circuit is not defined in the backend, so we can't change this

tests/integration/backend_test.py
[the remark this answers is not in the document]
cqc-melf: Should be solved now, I have removed this.

cqc-alec (approved):

PR #209, opened 2023-08-23
Controllable, predictable targeting of two-qubit gate type

cqc-alec:
* Hard-code a default two-qubit gate per device (currently `ZZPhase` for all), and document it.
* Allow users to choose a different target gate using new `*_with_options()` methods for compilation.
* By default, request that the backend does not alter the two-qubit gate of the submitted circuit.
* Allow users to override this request with a new option to `process_circuits()`.
* Add methods to query the backend for the default two-qubit gate and the set of supported two-qubit gates.

Integration tests ran successfully.


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -107,10 +108,14 @@
 }
 
 
+def _default_2q_gate(device_name: str) -> OpType:
+    # If we change this, we should update the main documentation page and highlight it
+    # in the changelog.
+    return OpType.ZZPhase
```
sjdilkes, line 114: I may be leaning towards a static default again, i.e. this, because:
• It sounds like a dynamic default may change as regularly as day to day
• The optimal gate is likely circuit dependent, meaning the dynamic default is likely inaccurate
• So, if dynamic, we're providing inconsistent, and possibly confusing, compilation for the sake of a metric that is likely not accurate
• This is assuming that `ZZPhase` is a reasonable default choice for a long (~many months) time
• This is also assuming fidelity information for operations becomes available (which I believe it is?)
• The documentation is clear regarding how to recompile, so savvy users should easily be able to do so


We should also consider looking at a new noise aware compilation method for these devices that considers circuit structure and gate fidelities when picking a target 2qb gate.

sjdilkes (changes requested):

isobelhooper:
Feedback on how this could affect Nexus:
* We would probably try to make any recompilation of the circuit to use a different 2-qubit gate happen when compiling in Nexus, rather than in L4.
  * Because Nexus doesn’t (and can’t, I believe) record any changes to the circuit that happen within L4, the user won’t know that the circuit that was actually run on the hardware wasn’t the same circuit we stored in Nexus, and we want to avoid that. So calls from Nexus would pass `allow_2q_gate_rebase=False` by default.
  * I’m not sure if Nexus would expose `allow_2q_gate_rebase` at all as an option in `process_circuits()` for running on Quantinuum hardware, for the same reason it doesn't expose `no_opt`. Would want to discuss with the Nexus team.
* Would it be reasonable to add an option to `QuantinuumBackend.get_compiled_circuit(s)` to check the reported default 2-qubit gate and set it just before compiling? I see you can pass it as an option to `default_compilation_pass_with_options`, but not to the function that does the compilation.
  * If not, we could potentially add it in Nexus to how we compile circuits for Quantinuum, by invoking `default_compilation_pass_with_options` ourselves if needed.

cqc-alec:
I think #210 is now my preferred approach.

Issue #232, opened 2023-09-13
enable wasm in qir tests after remote update


cqc-melf:
we have changed the wasm interface in qir, the tests needs to be enabled again after this is update on the remote side

cqc-alec:
```
FAILED integration/backend_test.py::test_wasm[Language.QIR-H1-1SC] - pytket.extensions.quantinuum.backends.quantinuum.GetResultFailed: Cannot retrieve result; job status is Circui...
```

Full error:

```
"1000: Compile error: 3 Encountered Unexpected Instruction   %2 = call i64 @add_one(i64 %1)"
```

cqc-melf:
Still not working, fails on production with:
`"1000: Compile error:  Invalid creg index for creg [\\"hyb_res0\\"]: 32"`

running wasm on production still shows the error as above

cqc-melf:
completed via #396

PR #233, opened 2023-09-13
update wasm, pytket and pyktet-qir

cqc-melf:
disable wasm test because of https://github.com/CQCL/pytket-quantinuum/issues/232
update pytket
update pytket-qir

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket >= 1.20.0rc2, <2.0",
+        "pytket ~= 1.20",
         "requests >= 2.2",
         "types-requests",
         "websockets >= 7.0",
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket ~= 1.19",
+        "pytket >= 1.20.0rc2, <2.0",
         "requests >= 2.2",
         "types-requests",
         "websockets >= 7.0",
--- a/tests/integration/backend_test.py
+++ b/tests/integration/backend_test.py
@@ -380,7 +380,7 @@ def test_classical(
     c.add_classicalexpbox_register(a - b, d.to_list())
 
     if language == Language.QASM:  # remove this when division supported in QIR
-        c.add_classicalexpbox_register(a * b // d, d)  # type: ignore
+        c.add_classicalexpbox_register(a * b // d, d.to_list())
 
     c.add_classicalexpbox_register(a << 1, a.to_list())
     c.add_classicalexpbox_register(a >> 1, b.to_list())
--- a/tests/integration/backend_test.py
+++ b/tests/integration/backend_test.py
@@ -382,8 +382,8 @@ def test_classical(
     if language == Language.QASM:  # remove this when division supported in QIR
         c.add_classicalexpbox_register(a * b // d, d)  # type: ignore
 
-    c.add_classicalexpbox_register(a << 1, a)  # type: ignore
-    c.add_classicalexpbox_register(a >> 1, b)  # type: ignore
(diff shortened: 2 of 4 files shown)


tests/integration/backend_test.py
```diff
@@ -376,14 +376,14 @@ def test_classical(
     c.add_c_setreg(23, a)
     c.add_c_copyreg(a, b)
 
-    c.add_classicalexpbox_register(a + b, d)
-    c.add_classicalexpbox_register(a - b, d)
+    c.add_classicalexpbox_register(a + b, d)  # type: ignore
+    c.add_classicalexpbox_register(a - b, d)  # type: ignore
```
cqc-alec, line 380:
```suggestion
    c.add_classicalexpbox_register(a + b, d.to_list())
    c.add_classicalexpbox_register(a - b, d.to_list())
```
Not sure how these slipped through the net before. I think I'd rather make them type-correct like this than add `type: ignore` comments, since the latter can mask actual bugs. I agree the old version is easier to read, but we can probably fix up the typing in pytket so that it is also accepted.

tests/integration/backend_test.py
```diff
@@ -376,14 +376,14 @@ def test_classical(
...
 
     if language == Language.QASM:  # remove this when division supported in QIR
-        c.add_classicalexpbox_register(a * b // d, d)
+        c.add_classicalexpbox_register(a * b // d, d)  # type: ignore
```
cqc-alec, line 383:
```suggestion
        c.add_classicalexpbox_register(a * b // d, d.to_list())
```

tests/integration/backend_test.py
```diff
@@ -376,14 +376,14 @@ def test_classical(
...
 
-    c.add_classicalexpbox_register(a << 1, a)
-    c.add_classicalexpbox_register(a >> 1, b)
+    c.add_classicalexpbox_register(a << 1, a)  # type: ignore
+    c.add_classicalexpbox_register(a >> 1, b)  # type: ignore
```
cqc-alec, line 386:
```suggestion
    c.add_classicalexpbox_register(a << 1, a.to_list())
    c.add_classicalexpbox_register(a >> 1, b.to_list())
```

setup.py
```diff
@@ -43,7 +43,7 @@
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket ~= 1.19",
+        "pytket >= 1.20.0rc2, <2.0",
```
cqc-alec, line 46: Why do we want to make this change?

cqc-alec (changes requested):

setup.py
cqc-melf: "pytket ~= 1.19" will not install the pre-release, I wanted to have that in here because it includes the wasm fix to make sure this still works with this extension

tests/integration/backend_test.py
cqc-melf: I was not sure if this might change the way this circuit is now handled. It might be possible that the left side has less or more bits than the register we are writing to?

tests/integration/backend_test.py
cqc-melf: I was not sure if this might change the way this circuit is now handled. It might be possible that the left side has less or more bits than the register we are writing to?

cqc-alec: > I was not sure if this might change the way this circuit is now handled. It might be possible that the left side has less or more bits than the register we are writing to?

It won't make any functional difference. `add_classical_expbox` just passes the whole register as a `std::vector<Bit>` to the binders.

setup.py
cqc-alec: Don't the tests get run on the latest pre-release anyway?

cqc-melf: Yes, but I wanted to do a release candidate for other people to test this

cqc-melf: Yes, I thought we might could do a pre release for other people to test, but that is not really needed. They can just install it locally.

cqc-melf, line 46:
```suggestion
        "pytket ~= 1.20",

cqc-melf:
This PR is waiting for the pytket 1.20 release

Issue #236, opened 2023-09-14
Test failure: `test_simulator[Language.QIR-H1-1E]`


cqc-alec:
Error message is:

```
"1000: Compile error: 1 Encountered Unexpected Instruction   %0 = call i1* @create_creg(i64 2)"
```

cqc-alec:
This is with pytket 1.20.0rc3, pytket-qir 0.2.1rc1.

cqc-melf:
Still showing up with pytket-qir 0.3.0.

cqc-alec:
QIR tests have been commented out: please search for `#236` and uncomment to ensure they pass (or open separate issues) before closing this one.

PR #240, opened 2023-09-14
update pytket-qir and pytket version

cqc-melf:


cqc-alec (changes requested):
Need to upgrade the version here:

https://github.com/CQCL/pytket-quantinuum/blob/98f900c922468c390b61679a3f626878e796e272/.github/workflows/build-test#L66

cqc-alec:
Are all the test failures QIR-related? Are they all versions of #236 or are there other issues here?

If there are some QIR tests that we expect to fail currently can we mark them as expected failures in pytest?

cqc-melf:
> Are all the test failures QIR-related? Are they all versions of #236 or are there other issues here?
> 
> If there are some QIR tests that we expect to fail currently can we mark them as expected failures in pytest?

I had seen this issues in the past and I thought that is was already solved. I am not sure why it is showing up again, looks like there has been some changes in the backend. I will bring this up later today.

I think it is the best solution to disable the tests for now and do the release to get the wasm fix out.

cqc-alec:
There seems to be one test failure (`test_cancel`) which I don't understand. Maybe worth rerunning in case it is an ephemeral thing?

cqc-melf:
> There seems to be one test failure (`test_cancel`) which I don't understand. Maybe worth rerunning in case it is an ephemeral thing?

Looks like it is gone after a rerun, I have not seen that in any local runs

cqc-alec (approved):

PR #243, opened 2023-09-19
Release/0.23.0

cqc-melf:


cqc-alec (approved):

cqc-melf:
I think we should we set the QA run as required for merges into main? Any objections? @cqc-alec

Issue #197, opened 2023-08-14
Allow persisent storage of authentication tokens


cqc-alec:
Currently each new Python process using pytket-quantinuum (and requiring authentication) needs the user to enter a password manually. This is sufficiently irritating that people work around it in various (more or less secure) ways involving persistent storage of the password. We should provide a supported method.

yao-cqc:
closed by #247

PR #247, opened 2023-10-09
Feature/persistent token storage

yao-cqc:
`QuantinuumConfigCredentialStorage` saves API tokens in the default pytket configuration file. Users don't need to login if the saved refresh token is still valid.

Example use

```python
backend = QuantinuumBackend(
        device_name="H1-SC",
        api_handler=QuantinuumAPI(QuantinuumConfigCredentialStorage()),
    )
```


tests/unit/api1_test.py
```diff
@@ -180,10 +185,21 @@ def test_custom_login_flow(
         ),
     )
 
+    backend_3 = QuantinuumBackend(
+        device_name=fake_device,
```
cqc-melf, line 189: Do we want to test the fake device here? Not sure I fully understood this testcase?

cqc-melf (commented):
One small question

tests/unit/api1_test.py
yao-cqc: The `fake_device` is just a dummy string from the `mock_machine_info` fixture, it doesn't matter because we catch all outward calls with `requests_mock`.

The testcase here is to make sure that the `login` endpoint is called three times (by backend_{1,2,3}) with `assert login_route.call_count == 3`.  `backend_4` shouldn't call `login` because `backend_3` has cached the authentication tokens in storage.

Hope this makes sense

cqc-melf: Yes, thank you! Sounds good

cqc-melf (approved):

cqc-alec:
@yao-cqc  Thanks! We can close #197 ?

PR #248, opened 2023-10-10
Add more WASM tests

cqc-alec:
These designed to be run on the emulator and test correctness. Circuits generate some random bits and compute a function on them. The second test includes persistent state. The third test uses a multi-valued wasm function.

I have run these tests successfully on my machine using the production H1-1E.


cqc-melf:
Currently looking into the CI issue, will review after that is solved.

tests/integration/testfile.wasm
cqc-melf, line 1: Have you changed this file? I think this was used in the example? Are we sure that is still working?

tests/wasm/README.md
```diff
@@ -0,0 +1,23 @@
+Create a C file containing
```
cqc-melf, line 1: We had all the wasm generation files in the tket repo so far, do we want to keep it that way?

Not sure what the best solution is, I think it would be good to test the extracted function signatures in the tket ci runs as well.

cqc-melf (commented):
I think the QA tests should work now, thee is just a temporary issue at the moment.

tests/integration/testfile.wasm
cqc-alec: Yes, I regenerated it as `add1.wasm` from `add1.c`. The file that was there contained several functions in addition to `add_one` that were not in `add1.c`, so must have been generated from a different source file. To tidy everything up I regenerated it, but didn't change anything in the tests (except for the name of the file).

tests/wasm/README.md
cqc-alec: I think it makes sense to have the source files here, as it allows easy editing and regeneration. Ideally maybe we'd just have the source files here and have the CI generate the wasm, but I think it's OK to have both in source control.

To an extent the function signatures are implicitly tested, since the calls to `add_wasm_to_reg()` would fail if they were incorrect.

tests/integration/testfile.wasm
cqc-melf: But that would mean that the file is not compatible with this example anymore? https://github.com/CQCL/pytket-quantinuum/blob/develop/examples/wasm_in_pytket.ipynb

(Not sure why the example is not tested on the CI)

tests/wasm/README.md
cqc-alec: Sorry, I don't understand what you mean?

cqc-melf: The code for the wasm parsing and signature checks of the functions is in the tket repo. I think it would be good to have a wasm file similar to the two added in the PR tested there as well. That would make it possible to catch issues earlier when we change parts of the code in tket.

tests/integration/testfile.wasm
cqc-alec: Hmm, OK, but I don't see where that notebook is getting its file from? We can add a copy of the old file there perhaps.

tests/wasm/README.md
cqc-alec: Ah, I understand. Yes, I agree.

tests/integration/testfile.wasm
cqc-alec: Oh dear. The test has indeed been failing because it can't find the file -- but this does not cause the workflow to fail: https://github.com/CQCL/pytket-quantinuum/actions/runs/6438882602/job/17485916107

cqc-alec: And other notebooks have been failing too.

cqc-alec: #249

cqc-alec:
I've put the old test file in the examples directory where it's used.

I don't know why the QA tests timed out.

tests/integration/backend_test.py
```diff
@@ -999,3 +999,178 @@ def test_scratch_removal(authenticated_quum_backend: QuantinuumBackend) -> None:
...
+
+@pytest.mark.skipif(skip_remote_tests, reason=REASON)
+@pytest.mark.parametrize(
+    "authenticated_quum_backend", [{"device_name": "H1-1E"}], indirect=True
+)
+@pytest.mark.parametrize(
+    "language",
+    [
+        Language.QASM,
+        # https://github.com/CQCL/pytket-quantinuum/issues/236
```
cqc-melf, line 1012: I would assume that this one also needs to wait for the wasm implementation (see https://github.com/CQCL/pytket-quantinuum/issues/232 )

cqc-melf (approved):

PR #254, opened 2023-10-12
Update backend_test.py

cqc-melf:
I don't see a file with the 2 in the name on develop at the moment?


cqc-alec (approved):
Yes that was my mix-up, thanks for spotting.

cqc-melf:
For me this works locally, I will not wait for the QA test, any objections? @cqc-alec

cqc-alec:
> For me this works locally, I will not wait for the QA test, any objections? @cqc-alec

Sure we can merge this.

PR #257, opened 2023-10-13
Fix table

### LINKED ISSUE #256: Default compilation table is now missing from the API docs
There is usually a table that shows the passes applied by different levels of optimisation.

It looks like this is no longer in the API docs.

Possibly due to a change in https://github.com/CQCL/pytket-quantinuum/pull/246 ? Not sure.

cqc-alec:
Fixes #256 .


CalMacCQ:
The table now shows for me in a local docs build

Thanks :)

CalMacCQ (approved):

PR #268, opened 2023-10-26
fix notebook checks

cqc-melf:


cqc-melf:
This PR make the checks work again, I think we would take a look if we want to submit to the emulator each time, of if we want to do something else maybe? see https://github.com/CQCL/pytket-quantinuum/issues/269

.github/workflows/check-examples.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         base: ${{ github.ref }}
         filters: |
           examples:
-            - 'examples/**'
+            - '{examples/**, .github/**}'
```
yao-cqc, line 33: Is the following '.github/workflows/check-examples.yml' covered by .github/**?

cqc-melf: I have cleaned that up in [4def276](https://github.com/CQCL/pytket-quantinuum/pull/268/commits/4def2761cfcb036437741d1168759f9e9dd4b301)

yao-cqc (approved):

PR #292, opened 2023-11-30
Organise pytket-quantinuum API docs

CalMacCQ:


.github/workflows/docs/conf.py
```diff
@@ -44,102 +45,3 @@
 }
 
 autodoc_member_order = "groupwise"
-
-# The following code is for resolving broken hyperlinks in the doc.
-
-import re
```
CalMacCQ, line 50: I think this is no longer needed.

@yao-cqc do you know differently?

yao-cqc: Are there still any broken hyperlinks in the docs?

CalMacCQ: Not that I know of. I have already deleted these lines in the extensions config file on the TKET site.

cqc-alec: It looks like the link to BasePass (return type of `QuantinuumBackend.default_compilation_pass()` is no longer working. So some of this may still be needed?

CalMacCQ:
I did try to fix this with `sphinx_autodoc_typehints` 

This cleaned up the `compilation_config` arg to a default value of `None`.

I couldn't find a way to tidy up the `api_handler` arg. There should be a way.

.github/workflows/docs/conf.py
CalMacCQ: Note that this config file is now only for local builds and the cqcl.github.io pages.

The tket-site only uses a single config file for the extensions. Maybe I should add these lines back in until I think of a better way to fix the links.

cqc-alec: It would be good to be able to do a local docs build that reproduces what will appear on the tket site, for checking changes before committing. Is that easy to do?

CalMacCQ: Okay. Perhaps what I should do is add the additonal `conf.py` lines back into this PR.

From my point of view I think it would be also fine to remove the cqcl.github.io site for each of the extensions as we have the tket.quantinuum.com site now.

It is a good point you raise about the configuratuion files diverging. I'll see if I can think of a way to be able to build the docs locally with the single `conf.py` file.

Note that you can already build the extensions docs locally on the tket-site repository. However this builds the docs from the content on the main branch (rather than develop). Maybe we should have a separate docs branch? I'm not sure what is best.

cqc-alec: > Note that you can already build the extensions docs locally on the tket-site repository. However this builds the docs from the content on the main branch (rather than develop). Maybe we should have a separate docs branch? I'm not sure what is best.

I do think having a long-lived `docs` branch is better than always building from `main`, though it wouldn't solve the local-builds problem since we want to be able to build locally from any branch. Could the script on the tket-site repo take a branch as an argument perhaps?

cqc-alec: > Okay. Perhaps what I should do is add the additonal `conf.py` lines back into this PR for now.

I think that's a good idea, while we work out the longer-term solution (which probably involves not publishing to github pages).

CalMacCQ: >  Could the script on the tket-site repo take a branch as an argument perhaps?

I was just thinking this. Worth a try I think.

CalMacCQ: I've put the deleted lines back in for now.

This won't fix the links on the tket wesite domain. Only on the cqcl.github.io site and the local builds.

cqc-alec (approved):

PR #297, opened 2023-12-07
Remove redundant type declarations from docstrings

CalMacCQ:
We now use the `sphinx-autodoc-typehints` extension. With this types can be inferred from the source code so we don't need to specify them twice anymore.

### DIFF
--- a/.github/workflows/docs/requirements.txt
+++ b/.github/workflows/docs/requirements.txt
@@ -1,4 +1,4 @@
-sphinx == 6.2.1
+sphinx ~= 6.2
 sphinx_book_theme >= 1.0.1, <2.0
 sphinx-autodoc-typehints
 sphinx-copybutton
--- a/.github/workflows/docs/requirements.txt
+++ b/.github/workflows/docs/requirements.txt
@@ -1,4 +1,4 @@
-sphinx >= 4.3.2, <6.2.0
+sphinx == 6.2.1
 sphinx_book_theme >= 1.0.1, <2.0
 sphinx-copybutton
 enum-tools[sphinx]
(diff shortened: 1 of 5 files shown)


.github/workflows/docs/requirements.txt
```diff
@@ -1,4 +1,5 @@
-sphinx >= 4.3.2, <6.2.0
+sphinx == 6.2.1
```
cqc-alec, line 1:
```suggestion
sphinx ~= 6.2
```
Do we need to pin to an exact version?

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -731,7 +699,8 @@ def process_circuits(
         """
         See :py:meth:`pytket.backends.Backend.process_circuits`.
 
-        Supported kwargs:
+        Supported kwargs
+        ================
```
cqc-alec, line 703: Can we make the text smaller by using a lower-level heading?

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -746,7 +715,7 @@ def process_circuits(
         * `pytketpass`: a ``pytket.passes.BasePass`` intended to be applied
            by the backend (beta feature, may be ignored).
         * `no_opt`: if true, requests that the backend perform no optimizations
-        * `allow_2q_gate_rebase`: if true, allow rebasing of the two-qubit gates to a
+        * `allow_2q_gate_rebase`: If true, allow rebasing of the two-qubit gates to a
```
cqc-alec, line 718: Prefer lower case for consistency.

cqc-alec (changes requested):

.github/workflows/docs/requirements.txt
CalMacCQ: Not really. I think our theme has to be <7 so I could change it to >= 4.3.2 <7 if preferred.

cqc-alec: No harm in requiring >= 6.2 though I think.

CalMacCQ: done

PR #298, opened 2023-12-08
Make `QuantinuumBackend.cost()` safe

### LINKED ISSUE #291: Remove `syntax_checker` argument to `QuantinuumBackend.cost()`
It is a footgun. Just have it automatically detected when possible, and error otherwise (which shouldn't happen).

cqc-alec:
Closes #291 .

(The argument is still needed for the emulators, but this ensures it won't accidentally incur costs.)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,14 @@
 Changelog
 ~~~~~~~~~
 
+Unreleased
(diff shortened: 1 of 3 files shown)


ss2165 (approved):
looks good, thanks

Issue #299, opened 2023-12-12
Circuits containing division operations not supported in QIR submissions


cqc-alec:
Under limited circumstances they are supported in QASM submissions, so this is a feature discrepancy.

cqc-alec:
Related: https://github.com/CQCL/pytket-qir/issues/104 and https://github.com/qir-alliance/pyqir/issues/165 .

cqc-alec:
Update: division operations with non-constant denominator shouldn't be supported at all, so it's actually a "bug" that they are in QASM.

cqc-melf:
We are not submitting anymore via qasm, I would like to close this. Please reopen if there is something left that could be done.

Issue #201, opened 2023-08-16
add support for wasm functions with more than one return value


cqc-melf:

cqc-melf:
From my understanding the code that had this problem is no longer in this repo. Please reopen if still needed.

PR #301, opened 2023-12-12
Remove not working testcases

cqc-melf:
remove testcases that should not work


cqc-melf:
Solves https://github.com/CQCL/pytket-quantinuum/issues/201 and https://github.com/CQCL/pytket-quantinuum/issues/299

tests/integration/backend_test.py
```diff
@@ -423,43 +423,6 @@ def test_classical(
...
-        pytest.param(
-            Language.QIR,
-            marks=pytest.mark.xfail(
-                reason="https://github.com/CQCL/pytket-quantinuum/issues/299"
-            ),
-        ),
-    ],
-)
-@pytest.mark.timeout(120)
-def test_division(
```
cqc-alec, line 445: Rather than delete these tests we could modify them so that they expect an exception to be raised; then swap the "xfail" annotation between QASM and QIR.

(I think we will want to support both these cases one day.)

cqc-melf: Good point, done in the added commits.

PR #302, opened 2023-12-13
Add local-emulator backends as optional feature

cqc-alec:
When pytket-quantinuum is installed with the `pecos` option (`pip install pytket-quantinuum[pecos]`), local-emulator backends become available.

For every hardware device detected as available, a local-emulator device is created, having the same name with "LE" appended. This device presents the same API as the actual device (with the exception of a few methods that are not available), but circuit execution is emulated (noiselessly) on the user's machine using PECOS.

Because of current limitations in pytket-phir, functionality is limited: circuits containing classical operations or WASM are not supported. Support for these will be added in due course. Noise emulation will also be added as and when the real devices' noise information becomes available over the API.

I've added a section to the documentation, making clear that this is still an experimental feature with limited functionality.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -148,7 +148,7 @@ jobs:
         ./.github/workflows/docs/check-build-docs
 
   qa_checks:
-    name: Run tests with QA endpoint and local simulator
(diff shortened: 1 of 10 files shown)


yao-cqc (approved):
Looks good to me!

PR #324, opened 2024-01-12
Update python support to 3.10, 3.11 and 3.12

sjdilkes:
This PR drops python 3.9 support and adds python 3.12.

# Checklist


cqc-alec (commented):
It looks like `llvmlite` does not yet have wheels available for Python 3.12. This is a test dependency. So we will have to disable the 3.12 tests on the CI until the next llvmlite release (just change 3.12 to 3.11 in the CI runs, and make a github issue to remind us to change it back when we can).

cqc-alec:
Also, pytket-pecos doesn't currently have 3.12 support (waiting for a pytket-phir release to enable this). This means that installing with the "pecos" option won't work with 3.12. We should add a note about this in the "Local Emulators" section in `docs/intro.txt`.

sjdilkes:
> It looks like `llvmlite` does not yet have wheels available for Python 3.12. This is a test dependency. So we will have to disable the 3.12 tests on the CI until the next llvmlite release (just change 3.12 to 3.11 in the CI runs, and make a github issue to remind us to change it back when we can).

issue added: https://github.com/CQCL/pytket-quantinuum/issues/327

sjdilkes:
> Also, pytket-pecos doesn't currently have 3.12 support (waiting for a pytket-phir release to enable this). This means that installing with the "pecos" option won't work with 3.12. We should add a note about this in the "Local Emulators" section in `docs/intro.txt`.

note added

cqc-alec (approved):

PR #349, opened 2024-01-29
Make pytket-qir a required dependency of pytket-quantinuum.

cqc-alec:
# Description

I forget why it wasn't already, but I don't see why it can't be now.

# Checklist


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -883,34 +878,17 @@ def process_circuits(
                     for name, count in Counter(bit.reg_name for bit in c0.bits).items():
                         for i in range(count):
                             results_selection.append((name, i))
-                    try:
```
cqc-melf, line 886: Should we still do a try and catch for possible errors and give more specific details that the conversion failed and they could try the qasm way if they want to?

cqc-melf (commented):
Looks good, do you want to add something to the changelog?

I think the reason that we have not added this was that we wanted to think if we want to move this over into pytket and have it there next to the qasm conversion? Has this decision been made?

cqc-alec:
> Looks good, do you want to add something to the changelog?

Yes, I should add something to the changelog.

> I think the reason that we have not added this was that we wanted to think if we want to move this over into pytket and have it there next to the qasm conversion? Has this decision been made?

Not really, but as it's not needed anywhere else (yet) I think it's fine to have the dependency here. If and when we decide we want QIR conversion in pytket it will be easy enough to add the dependency there and remove the explicit requirement from here.

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-alec: Currently the only message we print on error is "try installing pytket-qir", and that shouldn't be necessary now. Better error messages could be added in another PR, though the error raised by `pytket_to_qir()` should contain at least some useful information.

docs/changelog.rst
```diff
@@ -4,6 +4,11 @@ Changelog
 0.29.0 (January 2024)
```
cqc-melf, line 4:
```suggestion
0.30.0 (January 2024)

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-melf: Yes, but that error will not contain any hints pytket-quantinuum specific. Happy to do this in a separate PR.

cqc-melf (dismissed):

docs/changelog.rst
cqc-melf: Not sure if this is required, we will clean this up before doing a release anyway. Feel free to merge without this.

cqc-melf (approved):

PR #350, opened 2024-01-29
Add multithreading option for local emulators

### LINKED ISSUE #333: Add option to use multithreaded version of local emulator

cqc-alec:
# Description

Add a keyword argument to `process_circuits()` to enable multi-threaded emulation.

I have split the local-emulator tests into two files. This is because, for reasons that are mysterious to me, when I added the multi-threaded test at the bottom of the existing file, pytest hung when it reached this test and never returned. I assume there must be some bad interaction between pytest's own use of threads and the way multithreading is implemented in pecos.

# Related issues

Closes #333 .

# Checklist


sjdilkes (changes requested):
Cool! The only note I have is reagrding the file name `local_emulator1_test.py` - is `local_emulator1` descriptive in a manner I have missed, or could either the code in this file be added to the other local emulator test file, or the file renamed to `local_emulator_multithreading_test.py` or similar?

cqc-alec:
> Cool! The only note I have is reagrding the file name `local_emulator1_test.py` - is `local_emulator1` descriptive in a manner I have missed, or could either the code in this file be added to the other local emulator test file, or the file renamed to `local_emulator_multithreading_test.py` or similar?

Fair point. I copied the pattern from `api_test` and `api1_test`, but I will change it to something more descriptive as you suggest.

sjdilkes (approved):

PR #378, opened 2024-03-08
Add support for TK2 gate

cqc-alec:
# Description

Previously we were looking only for "Rxxyyzz" as the gate name, not "TK2".

Also improve the default compilation passes when targeting this gate.

# Checklist


cqc-alec:
There is one test failure (a timeout) which doesn't seem to be related to this change.

yao-cqc (approved):
Looks good

PR #387, opened 2024-03-15
hotfix/error-messages

irfankhan10:
Please summarise the changes.

Update error message to explain emulators and syntax-checkers are available 24/7.

### DIFF
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -532,7 +532,7 @@ class QuantinuumBackend(Backend):
                 "organization": l4_event.get("organization", "fairshare"),
             }
             calendar_data.append(event)
-        calendar_data.sort(key=lambda item: item["start-date"]) # type: ignore
+        calendar_data.sort(key=lambda item: item["start-date"])  # type: ignore
         return calendar_data
 
     @property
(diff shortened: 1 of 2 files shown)


mlk621-qtm:
I noted the calendar list not returns start/end dates as strings rather than datetime objects... I feel like those should stay as datetime objects as when a viz is added it will most likely need datetime objects and also think someone wanting to do their own viz would prefer that rather than a string as they'd probably need to convert it back to datetime object. I thought what you had before returning datetime object was good. It was just printing in the notebook to clearly see it if its not in UTC and make it pretyt was all but I thought it was good the actual variable itself was datetime object 

image

mlk621-qtm (dismissed):

irfankhan10:
@mlk621-qtm I accidently dismisssed your stale review. Please can you review again. I think all tests should pass now.

mlk621-qtm (dismissed):

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -451,7 +451,7 @@ def get_calendar(
         start_date: datetime.datetime,
         end_date: datetime.datetime,
         localise: bool = True,
-    ) -> List[Dict[str, str]]:
+    ) -> List[Dict[str, object]]:
```
cqc-alec, line 454: I think it is better to use `Any` here. See e.g. https://stackoverflow.com/questions/39817081/typing-any-vs-object

cqc-alec (approved):

PR #389, opened 2024-03-18
Feature/add calendar visualisation

irfankhan10:
# Description

Added a function to visualise calendar data using matplotlib + update to the queue visibility article

# Checklist


pytket/extensions/quantinuum/backends/calendar_visualisation.py
```diff
@@ -0,0 +1,141 @@
+"""Submodule providing calendar visualisation functionality"""
```
cqc-alec, line 1: Please add copyright notice.

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -452,14 +462,14 @@ def get_calendar(
         start_date: datetime.datetime,
         end_date: datetime.datetime,
         localise: bool = True,
-    ) -> List[Dict[str, str]]:
-        r"""Retrieves the Quantinuum H-Series operational calendar
+    ) -> List[Dict[str, object]]:
```
cqc-alec, line 465: `Any` instead of `object`?

setup.py
```diff
@@ -54,6 +54,7 @@
     ],
     extras_require={
         "pecos": ["pytket-pecos ~= 0.1.18"],
+        "calendar": ["matplotlib ~= 3.8.3", "pandas ~= 2.2.1"],
```
cqc-alec, line 57: Documentation about this functionality and how to enable it should be added to the front documentation page.

cqc-alec (changes requested):
Looks like there are some merge conflicts to resolve.

setup.py
irfankhan10: Added to the front page as requested

tests/integration/backend_test.py
```diff
@@ -1399,3 +1403,23 @@ def test_get_calendar_raises_error(
...
+
+
+@pytest.mark.skipif(skip_mpl_tests, reason=REASON_MPL)
+@pytest.mark.timeout(120)
+@pytest.mark.mpl_image_compare
+def test_view_calendar(authenticated_quum_handler: QuantinuumAPI) -> Any:
+    backend = QuantinuumBackend(
+        api_handler=authenticated_quum_handler, device_name="H1-1"
+    )
+    return backend.view_calendar(month=2, year=2024)
```
cqc-alec, line 1425: Do you know why this test is failing?

pytket/extensions/quantinuum/backends/calendar_visualisation.py
cqc-alec: Let's wait for confirmation that this is the right format/

tests/integration/backend_test.py
irfankhan10: yes, the data returned does not match the data in the png file. We need to use normal credentials. In other words, it should not be a  QA test

docs/intro.txt
```diff
@@ -15,6 +15,15 @@ and Windows. To install, run:
             The user will be prompted for their login credentials when making API calls such as calling the ``process_circuits`` method or querying ``backend_info``.
 
 
+``pytket-quantinuum`` provides calendar visualization capabilities. The ``calendar`` extra-install-argument must be specified to install matplotlib and pandas.
+
+::
+
+    pip install pytket-quantinuum["calendar"]
```
cqc-alec, line 22: I don't think we want quotes here?

irfankhan10: should be around pytket-quantinuum. I will change

cqc-alec: > should be around pytket-quantinuum. I will change

Not needed around pytket-quantinuum either.

irfankhan10: OK. In the past, I ran into issues not having the quotes, but maybe this is because of zsh shell.

cqc-alec (approved):

PR #394, opened 2024-03-21
Remove `no_opt` and `allow_2q_gate_rebase` options

cqc-alec:
# Description

Set both `no-opt` and `noreduce` to `True` to prevent any changes to the submitted circuit. It is assumed that the submitted circuit is what is desired to be run.

(It's still possible for these options to be overridden by providing an `options` dictionary to `process_circuits()` or `submit_program()`, but this is very much opt-in for those who know what they're doing.)

# Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,13 @@
 Changelog
 ~~~~~~~~~
 
+Unreleased
+----------
+
+* Remove ``no_opt`` and ``allow_2q_gate_rebase`` options to
+  ``process_circuits()`` and ``submit_program()``, and assume that the submitted
+  circuit is exactly what is desired to be run.
+
 0.31.0 (March 2024)
 -------------------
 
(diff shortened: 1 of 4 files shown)


cqc-alec:
There is a timeout on one of the tests, which I think we just have to ignore.

irfankhan10:
Hi Alec, this looks good. Is it worth setting the L4 option nativetq, or is that not needed?

cqc-alec:
> Hi Alec, this looks good. Is it worth setting the L4 option nativetq, or is that not needed?

Since we set `noreduce` to `True` by default I don't think it's needed. In the (presumably rare) case where the user wants `noreduce=False` but still wants to specify `nativetq`, they can still do so via the `options` argument.

irfankhan10:
Thanks Alec. 

Just want to clarify: 
- **no-opt**=True, means **tket-op-level** does not need to be specified?
- **no-reduce**=True means **nativetq** does not need to be specified?

cqc-alec:
> Just want to clarify:
> 
>     * **no-opt**=True, means **tket-op-level** does not need to be specified?
> 
>     * **no-reduce**=True means **nativetq** does not need to be specified?

That's my understanding yes.

irfankhan10 (approved):

PR #407, opened 2024-04-02
Ka/emulators

irfankhan10:
# Description

Knowledge article on Emulators. Statevector, Stabilizer, local emulation. Customisation of noise-model on cloud emulator.

# Checklist


yao-cqc (dismissed):

mlk621-qtm:
Is this the case? I thought there would be no noise so perfect emulation if the device had no noise? We had someone ask / test this as they were trying to do this 

"Only shot noise is present in the emulation result."

mlk621-qtm:
I'd like to use this style of HTML image in order to control the image size since it displays quite large in the notebook and put the image into the `figures` folder. 

```
<div style="text-align: center;">
         <img src="figures/emulator_usecase-2.png" width="800" />
</div>
```

mlk621-qtm:
Do the local emulators enable : Stabilizer emulation, noise model customization? this wasn't clear. I think I edited it so those are outlined for cloud emulators but if they're somehow relevant/available for local emulators maybe make that change.

mlk621-qtm:
in the figure here remove the hyphens for 32-qubits and 1000-qubits, Use-case to use case

To match the same tense/action (I'm forgetting my English grammar skills) with the rest of the bullets, I'd update "Exploring" to "Explore"

image

mlk621-qtm:
In these sections I didn't include descriptions of the noise model parameters and referred people to the data sheet they should look at (though I see I didn't mention H2 data sheet) in order to minimize needing to update emulator examples in future. We've already updated the parameters options at least twice, so ... nice to not to also need to update this. 

image

irfankhan10:
> Is this the case? I thought there would be no noise so perfect emulation if the device had no noise? We had someone ask / test this as they were trying to do this
> 
> "Only shot noise is present in the emulation result."

AFAIK, Yes. The pecos emulator run method necessarily requires n_shots to be specified as an integer. Not having a noise model means the device noise is not present. Sampling noise will still be present. I can verify this with the pecos people.

irfankhan10:
> Do the local emulators enable : Stabilizer emulation, noise model customization? this wasn't clear. I think I edited it so those are outlined for cloud emulators but if they're somehow relevant/available for local emulators maybe make that change.

Thanks Megan

1. The local emulators enable stabilizer emulation.
2. As things stand, right now with pytket-quantinuum the noise-model on the local emulator cannot be modified. 

I can update the pictures and specify this in the introduction and noiseless emulator section. Is there anywhere else this change should be made.

irfankhan10:
T

> In these sections I didn't include descriptions of the noise model parameters and referred people to the data sheet they should look at (though I see I didn't mention H2 data sheet) in order to minimize needing to update emulator examples in future. We've already updated the parameters options at least twice, so ... nice to not to also need to update this.
> 
> image

OK. I can remove this from here. Thanks for highlighting

1. FYI, this is not wasted work, since the details here can be recycled for the H-Series docs. 

2. Also by completing this exercise, I think the noise parameter definitions have become a little clearer. In the PDF, for certain categories of noise parameters, background details are given but not necessarily what each parameter means.

irfankhan10:
Thanks for reviewing, I can make those changes.

mlk621-qtm (approved):

mlk621-qtm (approved):

yao-cqc (approved):

PR #421, opened 2024-04-17
remove tket-opt-level from compilation example

irfankhan10:
# Description
remove tket-opt-level from compilation example notebook to streamline compilation workflow narrative.

# Checklist


mlk621-qtm (approved):
We may consider moving the contents of this to the regular circuit submission notebook now that the workflow has been streamlined.

irfankhan10:
I think this is a good point. There is also another example targeting compilation without Quantinuum API. Should all these compilation examples be collapsed into the submission notebook?

mlk621-qtm:
Maybe collapse it into the Quantinuum_circuit_submissions notebook? I think that is probably best place, just include basics of TKET compilation when introducing the user workflow, since there's nothing funky going on with submission anymore. 

The examples showing how to use QuantinuumBackend without having to sign in is mainly for MSFT users who wanted to try TKET compilation, but who don't have credentials on user portal. ... in my mind they are distinct for that reason since the rest of user workflow involving retrieving job results, etc. doesn't apply for those users. Thoughts?

mlk621-qtm (approved):

PR #435, opened 2024-05-16
Remove unreliable test.

cqc-alec:
I don't think it makes sense to have a test that fails if one happens to have `matplotlib` installed.


yao-cqc (approved):
Yeah, I agree

PR #458, opened 2024-06-17
Read max classical register width from backend info

cqc-alec:


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -1025,9 +1025,11 @@ def process_circuits(
                     quantinuum_circ = circuit_to_qasm_str(
                         c0,
                         header="hqslib1",
-                        maxwidth=32,
-                        # TODO Get maxwidth from self.backend_info once
-                        # https://github.com/CQCL/tket/issues/1445 fixed in pytket
+                        maxwidth=self.backend_info.misc.get(
```
cqc-melf, line 1028: Happy to copy this over to the other PR if you want this to be in there. See https://github.com/CQCL/pytket-quantinuum/pull/466

cqc-melf:
@cqc-alec can we close this?

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-alec: Forgot about this.

cqc-melf (approved):

PR #462, opened 2024-07-03
Don't use partial results.

cqc-alec:
This should cause no change to behaviour, but ensures that when partial results retrieval becomes the default we are not surprised by it.


cqc-melf (approved):
Should we disable the not working testcase?

cqc-alec:
> Should we disable the not working testcase?

We could mark it as xfail for now...

cqc-melf:
> > Should we disable the not working testcase?
> 
> We could mark it as xfail for now...

Happy to make a PR if you have not started one yet.

cqc-alec:
> > > Should we disable the not working testcase?
> > 
> > 
> > We could mark it as xfail for now...
> 
> Happy to make a PR if you have not started one yet.

Yes please, thanks.

Issue #457, opened 2024-06-14
compilation fails when OpType.ZZPhase is not in the gateset


cqc-melf:
When the defined default two qubit gate `OpType.ZZPhase` is not in the gatset of the device, the compilation fails.

We should add a check to test if `OpType.ZZPhase` is in the gateset, it is is not available throw and error or use something else.

PR #472, opened 2024-07-11
check if default two qubit gate is supported by the device

cqc-melf:
# Description

Solve https://github.com/CQCL/pytket-quantinuum/issues/457

# Checklist

### DIFF
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -1033,9 +1033,13 @@ class QuantinuumBackend(Backend):
                     quantinuum_circ = circuit_to_qasm_str(
                         c0,
                         header="hqslib1",
-                        maxwidth=32,
-                        # TODO Get maxwidth from self.backend_info once
-                        # https://github.com/CQCL/tket/issues/1445 fixed in pytket
+                        maxwidth=(
+                            self.backend_info.misc.get(
+                                "max_classical_register_width", 32
+                            )
+                            if self.backend_info
+                            else 32
+                        ),
                     )
                     used_scratch_regs = _used_scratch_registers(quantinuum_circ)
                     for name, count in Counter(
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -639,7 +639,6 @@ class QuantinuumBackend(Backend):
     @property
     def default_two_qubit_gate(self) -> OpType:
         """Returns the default two-qubit gate for the device."""
-        assert self._default_2q_gate in self.two_qubit_gate_set
         return self._default_2q_gate
 
     @property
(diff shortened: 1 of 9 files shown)


yao-cqc:
Can we instead check this in the `QuantinuumBackend` constructor and use the same error message in `set_compilation_config_target_2qb_gate`?

cqc-melf:
> `set_compilation_config_target_2qb_gate`

`set_compilation_config_target_2qb_gate` already has a check, so I think that should be fine?

Unfortunately I can't use _gate_set in the constructor because that calls the API and breaks quiet a few things for backends constructed with the offlineapi.

Happy to find a workaround for that if you think that is important.

cqc-alec:
It does feel wrong to set this config value inside rebase_pass(). Can we change the `_default_2q_gate()` function instead?

cqc-melf:
> By the way, `deadhead` and `brathead` both now accept the `ZZPhase` gate. So maybe we don't need this anymore, as `ZZPhase` is available on all devices?

I think we discussed this in the last sprint planing that we still want to have this check additionally to make this future proof.  

If you think this is not required, then we can close this.

cqc-melf:
> > > `set_compilation_config_target_2qb_gate`
> > 
> > 
> > `set_compilation_config_target_2qb_gate` already has a check, so I think that should be fine?
> > Unfortunately I can't use _gate_set in the constructor because that calls the API and breaks quiet a few things for backends constructed with the offlineapi.
> > Happy to find a workaround for that if you think that is important.
> 
> I think `self._gate_set` won't make any API call if the backend is constructed with the offline API. Maybe I misunderstood the problem?

I think I was wrong with the problem detection, the actual issue is that we have tests are not expected to make API calls that would make this calls with the changes in the constructor.

I have now gone the way and updated all the affected testcases.

cqc-alec (commented):
I'm just a little worried that requiring login when constructing `QuantinuumBackend` may break some workflows; perhaps we should check this with nexus?

cqc-melf:
> I'm just a little worried that requiring login when constructing `QuantinuumBackend` may break some workflows; perhaps we should check this with nexus?

I checked with them, this is not a concern.

cqc-melf:
I will take a look at the failed notebook and update that. Will ask for a review after that is updated.

cqc-alec:
> I will take a look at the failed notebook and update that. Will ask for a review after that is updated.

This is probably #474 . If so it needs an update from the US side.

cqc-alec (dismissed):

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -1038,7 +1038,7 @@ def process_circuits(
                                 "max_classical_register_width", 32
```
cqc-alec: If we want to change the default we should change it here too, but I'm not sure if we want to do this.

cqc-alec (approved):

PR #477, opened 2024-07-18
Update to pytket-pecos 0.1.29

cqc-alec:


setup.py
```diff
@@ -54,7 +54,7 @@
         "numpy < 2",
     ],
     extras_require={
-        "pecos": ["pytket-pecos ~= 0.1.28"],
+        "pecos": ["pytket-pecos ~= 0.1.29"],
```
cqc-melf, line 57: do we want this to be `>=` ?

cqc-alec: I thought about this, but decided that we would probably only bump pytket-pecos to 0.2 when there's a change that would break pytket-quantinuum, so it's better to keep it this way for now.

cqc-melf (approved):

PR #478, opened 2024-07-18
Skip test needing access to authenticated backend when not running remote tests

cqc-alec:
Necessary since #472 requires authentication to construct `QuantinuumBackend`.

### DIFF
--- a/tests/integration/backend_test.py
+++ b/tests/integration/backend_test.py
@@ -75,16 +75,14 @@ REASON_MPL = "PYTKET_RUN_MPL_TESTS not set \
 (requires configuration of Quantinuum username)"
 
 
+@pytest.mark.skipif(skip_remote_tests, reason=REASON)
 @pytest.mark.parametrize("authenticated_quum_backend_qa", [None], indirect=True)
 @pytest.mark.parametrize("language", [Language.QASM, Language.QIR])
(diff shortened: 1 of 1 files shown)


tests/integration/backend_test.py
```diff
@@ -75,16 +75,14 @@
...
 
+@pytest.mark.skipif(skip_remote_tests, reason=REASON)
 @pytest.mark.parametrize("authenticated_quum_backend_qa", [None], indirect=True)
 @pytest.mark.parametrize("language", [Language.QASM, Language.QIR])
 @pytest.mark.timeout(120)
 def test_quantinuum(
     authenticated_quum_backend_qa: QuantinuumBackend, language: Language
 ) -> None:
-    if skip_remote_tests:
-        backend = QuantinuumBackend(device_name="H1-1SC", machine_debug=True)
```
cqc-melf, line 85: We could put in the offline api here?

cqc-alec: The problem seems to be that the test is parametrized by a "authenticated_quum_backend_qa", which I suppose gets evaluated even if `skip_remote_tests` is `True`. So just changing this line doesn't help.

cqc-melf:
the timeout is somehow expected, so feel free to merge without the test working.

PR #482, opened 2024-08-20
Update device info in offline API and in some tests

cqc-alec:
Brings the backend info up to date with current reality.

Also remove a test for constructing backends using a device family (like "H1"), which doesn't work any more.


cqc-alec:
CI tests failing because of an unrelated credentials issue.

yao-cqc (approved):
Looks good!

Issue #504, opened 2024-10-09
Use API handler for available devices requests


johnchildren:
There is some slightly surprising behavior where the QuantinuumBackend class will make requests directly for `_available_devices`  rather than calling a QuantinuumAPI method here: https://github.com/CQCL/pytket-quantinuum/blob/642243ce72330e95e42f90cbcc8920b49f31a7f9/pytket/extensions/quantinuum/backends/quantinuum.py#L373

It would be easier to write alternative QuantinuumAPI implementations if this request was handled by the api_handler itself.

PR #505, opened 2024-10-10
feat: Lazily load 2qb gate

johnchildren:
# Description

Rather than fetching the 2qb gate of devices on backend construction, this commit changes the behavior of the backend to only fetch devices once the 2qb gate is requested and caches the value as a cached_property using a decorator. As a side effect it removes the need for the user to log in to construct a backend.

Changes to tests that were required due to the backend now requiring credentials to construct have been reverted.

# Related issues

Tangentially related to https://github.com/CQCL/pytket-quantinuum/issues/504

# Checklist


yao-cqc (approved):
Looks good!

Issue #510, opened 2024-10-11
Incorrect shots length after submitting QIR


cqc-alec:
When running the CI tests after switching the default submission language to QIR:

```
FAILED integration/backend_test.py::test_scratch_removal[H1-1SC] - assert False
 +  where False = all(<generator object test_scratch_removal.<locals>.<genexpr> at 0x7f7fec9bb5a0>)
```

See #508 .

cqc-melf:
I don't see a reason why this should only be tested with the default language, any objections on updating this?

cqc-melf:
The circuit has the following bits:
```[cb0[0], cb0[1], cb1[0], cb1[1], cb1[2], tk_SCRATCH_BIT[0], tk_SCRATCH_BIT[1], tk_SCRATCH_BIT[2]]```
We get all 8 of them back, I would assume that we want to exclude the `tk_SCRATCH_BIT` at some point. Do we want to do this when generating the qir or later when we get the results?

**Comment by cqc-melf** [comment]
So I think we are missing the part here which is only done for qasm:
```
                    used_scratch_regs = _used_scratch_registers(quantinuum_circ)
                    for name, count in Counter(
                        bit.reg_name
                        for bit in c0.bits
                        if not _is_scratch(bit) or bit.reg_name in used_scratch_regs
                    ).items():
                        for i in range(count):
                            results_selection.append((name, i))

```

I will make a PR to add this

Issue #509, opened 2024-10-11
`AttributeError: 'QirGenerator' object has no attribute 'wasm'`


cqc-alec:
When running the CI tests after switching the default submission language to QIR:

```
FAILED integration/backend_test.py::test_wasm_costs[H1-1E] - AttributeError: 'QirGenerator' object has no attribute 'wasm'
```

See #508 .

cqc-melf:
Do we want to add an language option to the cost function? Or is that always using the default?

cqc-alec:
> Do we want to add an language option to the cost function? Or is that always using the default?

We should probably add the option to the `cost()` function, with the same default as `process_circuits()`. I'd expect the costs to be the same but you never know...

PR #512, opened 2024-10-11
fix issues with QIR in cost function and update expected results

cqc-melf:
# Description

Please summarise the changes.

# Related issues

Solves https://github.com/CQCL/pytket-quantinuum/issues/510
and Solves https://github.com/CQCL/pytket-quantinuum/issues/509 

# Checklist

### DIFF
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -1451,10 +1451,17 @@ jobid is {jobid}"
             raise ValueError(f"Device {backend._device_name} is not a syntax checker.")
 
         wasm_fh = cast(Optional[WasmFileHandler], kwargs.get("wasm_file_handler"))
+        language: Optional[Language] = cast(
+            Language, kwargs.get("language"), Language.QASM
+        )
 
         try:
             handle = backend.process_circuit(
-                circuit, n_shots, kwargs=kwargs, wasm_file_handler=wasm_fh
+                circuit,
+                n_shots,
+                kwargs=kwargs,
+                wasm_file_handler=wasm_fh,
+                language=language,
             )
         except DeviceNotAvailable as e:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -1078,6 +1076,12 @@ def process_circuits(
                             ),
                         )
                     ).decode("utf-8")
+
+                    for name, count in Counter(
```
cqc-melf, line 1080: qasm only excludes unused scratch bits, I have now decided to exclude all of them here for QIR. I am not sure if that is the right thing to do, but I would assume that no users are interested in seeing this?
Open to discussions.

tests/integration/backend_test.py
```diff
@@ -1106,12 +1106,22 @@ def test_old_handle(
     assert (r0.get_shots() == r1.get_shots()).all()
 
 
+@pytest.mark.parametrize(
```
cqc-alec, line 1109: Should we parametrize the other failing test as well?

cqc-alec (approved):
I rebased #508 onto this branch and the only failure was a timeout, so happy with the fixes.

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-alec: Maybe QASM should exclude all of them as well?

tests/integration/backend_test.py
cqc-melf: done

tests/integration/backend_test.py
```diff
@@ -834,9 +834,18 @@ def test_wasm(
...
+    [
+        Language.QASM,
+        Language.QIR,
+        Language.PQIR,
+    ],
+)
 @pytest.mark.timeout(120)
 def test_wasm_costs(
     authenticated_quum_backend_qa: QuantinuumBackend,
+    language: Language,
```
cqc-alec, line 848: Should use this in the call to `cost()`?

cqc-melf: Yes, done

cqc-alec (approved):

PR #522, opened 2024-10-23
docs: port docs source to markdown, fix links

CalMacCQ:
# Description

I've ported the docs source to MyST markdown rather than rst. This has two benefits.

1) Improved crossreferncing - MyST is better at resolving links than standard rst. I can replace all of the full URL links in the default compilation pass table with cross refernces to the docs. This makes us a lot less suseptable to URL breakage.
2) Means we can have code snippets that are stylyistically conistent with the rest of the docs.

Markdown is also just nice to write in.

Screenshot 2024-10-23 at 15 15 39


cqc-alec (approved):
Thanks!

PR #525, opened 2024-10-25
Set number of qubits for leakage detection

sjdilkes:
# Description

The automatic leakage detection tries to maximise parallelism by using as many spare device qubits as possible for performing leakage detection. Currently when passing `leakage_detection=True` to `process_circuit/s`, it automatically assumes that the user wants to make use of all qubits. This PR adds a new kwarg to `pytket-quantinuum` `n_leakage_detection_qubits` that allows the user to specify fewer qubits. 

# Related issues

Please mention any github issues addressed by this PR.

# Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -8,6 +8,7 @@
 
 - Use QIR by default for program submission.
 - Update pytket-qir version requirement to 0.15.
+- Add new `kwarg` `n_leakage_detection_qubits` to `process_circuits()`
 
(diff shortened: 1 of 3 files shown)


yao-cqc:
I think we should update the changelog as well.

sjdilkes:
> I think we should update the changelog as well.

Just done!

yao-cqc (approved):
Thanks!

Issue #501, opened 2024-10-02
Unclear type of result for QIR


cqc-melf:
Currently the type of the result is depended on the QIR submitted. For QIR with the register functions the result type is a bit list (as a string), for QIR without them the result type is a integer (as a string).

Example:
When I record an integer that has been generated by a call of get_int_from_creg I get a result back as a bit string.
`%9 = call i64 @get_int_from_creg(i1* %2)`
`call void @__quantum__rt__int_record_output(i64 %9, i8* getelementptr inbounds ([2 x i8], [2 x i8]* @2, i32 0, i32 0))`
Result:
```
{'a': ['00111001', '10011100', '00101010', '00001100', '10101100', '10110000', '00001001', '10110001', '01111010', '10100110'], 'b': ['0100', '0100', '0011', '0010', '0100', '0011', '0010', '0100', '0101', '0100'], 's': ['0', '1', '0', '0', '1', '1', '0', '1', '0', '1']}
```
When I record an integer that has been generated in a different way, I the results as an int.
`%136 = call i64 @get_c()`
`call void @__quantum__rt__int_record_output(i64 %136, i8* getelementptr inbounds ([2 x i8], [2 x i8]* @1, i32 0, i32 0))`
Result:
```
{'b': ['4', '4', '4', '6', '5', '5', '4', '4', '6', '5'], 'a': ['92', '232', '178', '231', '122', '203', '114', '149', '125', '110'], 's': ['0', '1', '1', '1', '0', '1', '0', '1', '0', '0']}
```

Currently discussing on how to solve this.

cqc-melf:
From my understanding the code that had this problem is no longer in this repo. Please reopen if still needed.

PR #528, opened 2024-10-30
Melf/add negative results handling

cqc-melf:
# Description

Please summarise the changes.

# Related issues

Please mention any github issues addressed by this PR.

# Checklist

### DIFF
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -1519,10 +1519,7 @@ This could indicate a problem with the circuit submitted"
             )
 
     if results_selection is None:
-        found_int_res = False
-        for creg, reslist in resultdict.items():
-            if any([re.findall("[23456789]", res) for res in reslist]):
-                found_int_res = True
+        found_int_res = any(re.findall("[23456789]", res) for reslist in resultdict.values() for res in reslist)
 
         if found_int_res:
             # this is only a temporary solution and not fully working
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -1513,8 +1513,7 @@ def _convert_result(
     if results_selection is None:
         for creg in resultdict:
             if any(["-" in res for res in resultdict[creg]]):
-                resultdict[creg] = [res.replace("-", "0") for res in resultdict[creg]]
-                warnings.warn(
+                raise ValueError(
                     f"found negative value for creg: {creg}. \
 This could indicate a problem with the circuit submitted"
                 )
@@ -1525,6 +1524,8 @@ This could indicate a problem with the circuit submitted"
                     found_int_res = True
 
         if found_int_res:
+            # this is only a temporary solution and not fully working
+            # see issue https://github.com/CQCL/pytket-quantinuum/issues/501
 
             def conv_int(res: str) -> list:
                 long_res = bin(int(res)).replace(
@@ -1571,27 +1572,20 @@ This could indicate a problem with the circuit submitted"
         c_bits = [Bit(name, ind) for name, ind in results_selection]
 
         # Construct the shots table
-        result_int = False
         try:
             stacked_array = [
                 [int(resultdict[name][i][-1 - ind]) for name, ind in results_selection]
                 for i in range(n_shots)
             ]
         except ValueError:
-            # this is only a temporary solution and not fully working
-            # see issue https://github.com/CQCL/pytket-quantinuum/issues/501
-            warnings.warn(
+            raise ValueError(
                 "found unexpected character in the result values. \
 This could indicate a problem with the circuit submitted"
             )
-            result_int = True
 
         except IndexError:
             # this is only a temporary solution and not fully working
             # see issue https://github.com/CQCL/pytket-quantinuum/issues/501
-            result_int = True
-
-        if result_int:
             stacked_array = [
                 [
                     int(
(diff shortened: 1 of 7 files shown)


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -1511,10 +1511,39 @@ def _convert_result(
...
-        }
+        for creg in resultdict:
+            if any(["-" in res for res in resultdict[creg]]):
+                resultdict[creg] = [res.replace("-", "0") for res in resultdict[creg]]
+                warnings.warn(
+                    f"found negative value for creg: {creg}. \
+This could indicate a problem with the circuit submitted"
+                )
+        found_int_res = False
+        for x in [f"{s}" for s in range(2, 10)]:
```
cqc-alec, line 1522: If an int result happens to consist only of 0s and 1s this will be missed?

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -1511,10 +1511,39 @@ def _convert_result(
...
+            for creg in resultdict:
```
cqc-alec, line 1523: Can we do this in a single loop with a regex?

cqc-alec (commented):
It seems we're just mangling results that we don't understand. What if we just raise an exception if the format is not what we expect?

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-melf: Yes, we need to wait for the results to contain a tag for this. See https://github.com/CQCL/pytket-quantinuum/issues/501

tests/integration/backend_test.py
```diff
@@ -1080,6 +1080,126 @@ def test_qir_submission_mz_to_reg_qa(
...
+    r = b.get_result(h)
+    assert len(r.get_shots()) == 10
+    # 192 = 3 * 64 , 102 = 64 + 32 + 6
+    assert len(r.get_bitlist()) in [102, 192]
+
+
+@pytest.mark.parametrize(
+    "language",
+    [
+        Language.QIR,
```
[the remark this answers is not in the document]
cqc-melf: Yes, added now

tests/integration/backend_test.py
```diff
@@ -1080,6 +1080,126 @@ def test_qir_submission_mz_to_reg_qa(
...
+
+    r = b.get_result(h)
+    assert len(r.get_shots()) == 10
+    assert len(r.get_bitlist()) == 102
+
+
+@pytest.mark.parametrize(
+    "language",
+    [
+        Language.QIR,
```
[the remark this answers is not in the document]
cqc-melf: Yes, added now

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-melf: Yes, done

tests/integration/backend_test.py
```diff
@@ -1080,6 +1080,158 @@ def test_qir_submission_mz_to_reg_qa(
...
+    c.Measure(Qubit(2), Bit(2))
+    c0 = c.add_c_register("c0", 63)
+    c1 = c.add_c_register("c1", 32)
+    c.add_c_setbits([False] * 29 + [True, True, True, True] + [False] * 30, list(c0))
+    c.add_wasm_to_reg("add_one", wfh, [c0], [c1])
+
+    wfh.check()
+
+    h = b.process_circuit(c, n_shots=10, language=language, wasm_file_handler=wfh)
+    with pytest.raises(ValueError):
```
cqc-alec, line 1231: Perhaps a comment explaining why we expect this?

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -1510,11 +1510,42 @@ def _convert_result(
...
+
     if results_selection is None:
-        array_dict = {
-            creg: np.array([list(a) for a in reslist]).astype(np.uint8)
-            for creg, reslist in resultdict.items()
-        }
+        found_int_res = False
+        for creg, reslist in resultdict.items():
+            if any([re.findall("[23456789]", res) for res in reslist]):
+                found_int_res = True
```
cqc-alec, line 1525:
```suggestion
        found_int_res = any(re.findall("[23456789]", res) for reslist in resultdict.values() for res in reslist)
```
Slightly more efficient because of shortcutting.

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -1510,11 +1510,42 @@ def _convert_result(
...
+
+        if found_int_res:
+            # this is only a temporary solution and not fully working
```
cqc-alec, line 1528: This isn't reliable; we could still miss integer results in some cases (e.g. "10").
Is there some way we can know what to expect when we submit the circuit, and store this information in the `ResultHandle`, instead of guessing from the results?

Do you know when we can expect a fix for this in L4?

cqc-melf: I agree, that is why I labeled this as a `temporary solution`. We could parse the QIR given by the user and try to predict it? We don't have the circuit, so we can't do any predictions based on the number of bits.
I am not sure when the fix is going to happen.

tests/integration/backend_test.py
cqc-melf: done, see [5e162bd](https://github.com/CQCL/pytket-quantinuum/pull/528/commits/5e162bd177cc6757063a7e88e525d02ba592dfc5)

cqc-alec (approved):

PR #541, opened 2024-11-08
fix: Use API handler for available devices

johnchildren:
# Description

Fixes the usage of the api handler in the available_devices method by re-using the session of the real handler rather than making a fresh session in the backend.

This will also make it easier to write alternative QuantinuumAPI implementations as a side effect.

# Related issues

Resolves https://github.com/CQCL/pytket-quantinuum/issues/504

# Checklist

### DIFF
--- a/pytket/extensions/quantinuum/backends/api_wrappers.py
+++ b/pytket/extensions/quantinuum/backends/api_wrappers.py
@@ -514,7 +514,7 @@ class QuantinuumAPI:
         return jr
 
     def get_machine_list(self) -> list[dict[str, Any]]:
-        """returns the given list of the available machines
+        """Returns a given list of the available machines
         :return: list of machines
         """
         id_token = self.login()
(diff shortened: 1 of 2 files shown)


pytket/extensions/quantinuum/backends/api_wrappers.py
```diff
@@ -513,6 +513,20 @@ def get_calendar(self, start_date: str, end_date: str) -> list[dict[str, str]]:
         jr: list[dict[str, str]] = res.json()
         return jr
 
+    def get_machine_list(self) -> list[dict[str, Any]]:
+        """returns the given list of the available machines
```
isobelhooper, line 517:
```suggestion
        """Returns a list of the available machines
```
(not vital at all as a change, just the inconsistent lowercase from the original was bugging me)

pytket/extensions/quantinuum/backends/api_wrappers.py
```diff
@@ -616,8 +630,8 @@ def __init__(self, machine_list: Optional[list] = None):
         self._cred_store = None
         self.submitted: list = []
 
-    def _get_machine_list(self) -> Optional[list]:
-        """returns the given list of the avilable machines
+    def get_machine_list(self) -> list[dict[str, Any]]:
+        """returns the given list of the available machines
```
isobelhooper, line 634:
```suggestion
        """Returns a list of the available machines
```
(also not vital, but while we're here...)

PR #546, opened 2024-11-21
Add `optimisation_level` 3 using `GreedyPauliSimp`

sjdilkes:
Adds a new optimisation level "3" that uses `GreedyPauliSimp` targeting `zzphase` to optimise circuits. This calls `RemoveBarriers`, which is part of the motivation for adding it as a new level of optimisation. This mimics the changes to `pytket-qiskit`.

# Checklist


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -726,6 +730,29 @@ def default_compilation_pass(self, optimisation_level: int = 2) -> BasePass:
                     RemoveRedundancies(),
                 ]
             )
+        else:
+            passlist.extend(
+                [
+                    RemoveBarriers(),
+                    AutoRebase({OpType.CX, OpType.Rz, OpType.H}),
```
yao-cqc, line 737: I think we should allow more gates that `GreedyPauliSimp` can handle so the rebase can run faster. Also easier for the Pauli graph conversion.

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -655,14 +657,16 @@ def rebase_pass(self) -> BasePass:
             allow_swaps=self.compilation_config.allow_implicit_swaps,
         )
 
-    def default_compilation_pass(self, optimisation_level: int = 2) -> BasePass:
+    def default_compilation_pass(
+        self, optimisation_level: int = 2, timeout: int = 300
+    ) -> BasePass:
         """
         :param optimisation_level: Allows values of 0,1 or 2, with higher values
```
yao-cqc, line 664:
```suggestion
        :param optimisation_level: Allows values of 0,1,2 or 3, with higher values
```

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -726,6 +730,29 @@ def default_compilation_pass(self, optimisation_level: int = 2) -> BasePass:
...
+        else:
+            passlist.extend(
+                [
+                    RemoveBarriers(),
+                    AutoRebase({OpType.CX, OpType.Rz, OpType.H}),
+                    GreedyPauliSimp(
+                        allow_zzphase=True,
+                        only_reduce=True,
+                        thread_timeout=timeout,
+                        trials=10,
```
yao-cqc, line 742: When `GreedyPauliSimp` returns false, will the user know the cause, so they can, for example, try increasing the timeout?

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -726,6 +730,29 @@ def default_compilation_pass(self, optimisation_level: int = 2) -> BasePass:
                     RemoveRedundancies(),
                 ]
             )
+        else:
+            passlist.extend(
+                [
+                    RemoveBarriers(),
```
yao-cqc, line 736: Should we also issue a warning if the circuit contains barriers?

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -655,14 +657,16 @@ def rebase_pass(self) -> BasePass:
             allow_swaps=self.compilation_config.allow_implicit_swaps,
         )
 
-    def default_compilation_pass(self, optimisation_level: int = 2) -> BasePass:
+    def default_compilation_pass(
```
yao-cqc, line 660: should also document the `timeout` parameter

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -655,14 +657,16 @@ def rebase_pass(self) -> BasePass:
...
+        self, optimisation_level: int = 2, timeout: int = 300
```
yao-cqc, line 661: It seems the user cannot set a `timeout` when calling `get_compiled_circuit`. Any thoughts on this?

sjdilkes: I've added `get_compiled_circuit` and `get_compiled_circuits` to `QuantinuumBackend` with a timeout parameter

pytket/extensions/quantinuum/backends/quantinuum.py
sjdilkes: done

pytket/extensions/quantinuum/backends/quantinuum.py
sjdilkes: good point, added to the new overloaded `get_compiled_circuit` method

pytket/extensions/quantinuum/backends/quantinuum.py
sjdilkes: They won't & I can't think of a good way of sharing this information in the code. Might just have to make it clear in the documentation?

pytket/extensions/quantinuum/backends/quantinuum.py
sjdilkes: done

pytket/extensions/quantinuum/backends/quantinuum.py
yao-cqc: I see. Can we print out some information in the C++ code? If you think that's possible, we should create an issue for this.

yao-cqc (approved):

PR #547, opened 2024-11-21
docs: fix invalid table in docs

CalMacCQ:
# Description

Minor fix. After #546 the table had an invalid number of columns and the `RemoveBarriers` pass didn't show up. Should be fixed now.

Fixed the pytket-qiskit one in https://github.com/CQCL/pytket-qiskit/pull/421

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -90,7 +90,7 @@ a circuit, while optimisation level 3 will. At optimisation level 3 the default 
 the circuit 2-qubit gate count is not reduced after compilation.
 
 :::{list-table} **Default compilation pass for the QuantinuumBackend**
-:widths: 25 25 25
+:widths: 25 25 25 25
 :header-rows: 1
 
 * - optimisation_level = 0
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -104,7 +104,7 @@ the circuit 2-qubit gate count is not reduced after compilation.
 * - [AutoRebase [2]](inv:#*.AutoRebase)
   - [SynthesiseTket](inv:#*.SynthesiseTket)
   - [FullPeepholeOptimise [3]](inv:#*.passes.FullPeepholeOptimise)
-  - [RemoveBarriers](inv:#*.passes.RemoveBarriers)
```
CalMacCQ, line 107: Not sure why this first reference didn't resolve. Tested that this replacement works locally though.

cqc-alec: I don't know why it wouldn't work. Does this now show the name `_tket` in the docs? I think we want to avoid that.

CalMacCQ: Yeah unfortunately I'll need to fix that in the TKET API docs. There's a replacement of all of the `_tket` names that isn't working currently. However I would've thought this resolver used module imports and that the module names on the webpage wouldn't matter.

cqc-alec (approved):

PR #560, opened 2025-01-09
Add CliffordCircuitPredicate if we're using the stabilizer simulator

isobelhooper:
# Description

QuantinuumBackend currently doesn't check if a circuit compiled for or run on the stabilizer simulator is a Clifford circuit.  This PR fixes that by adding a predicate check in that case.

# Checklist


pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -615,6 +616,8 @@ def required_predicates(self) -> list[Predicate]:
             assert self.backend_info is not None
             preds.append(MaxNQubitsPredicate(self.backend_info.n_nodes))
             preds.append(MaxNClRegPredicate(cast(int, self.backend_info.n_cl_reg)))
+        if self.simulator_type == "stabilizer":
```
cqc-alec, line 619: Can we add a test or two showing that this is working?

cqc-alec: There's a related test [here](https://github.com/CQCL/pytket-quantinuum/blob/42cd71c7d581c6c94bd3cde759673ce73c18a3a1/tests/integration/backend_test.py#L599) which could be extended. (In fact, don't we expect that test to now throw an error from `process_circuit()`?)

isobelhooper:
Integration test is currently failing because `CliffordCircuitPredicate` doesn't allow `Measure` operations. It's reasonable for a stabilizer simulator to allow measurement, so this is a problem in the predicate itself. Alec's opened https://github.com/CQCL/tket/issues/1742 to address that, and this PR will wait until that's in TKET.

CalMacCQ:
pytket v1.40 is out now so I think this should be unblocked hopefully.

https://docs.quantinuum.com/tket/api-docs/changelog.html

cqc-alec (approved):
Thanks!

PR #570, opened 2025-01-28
Constrain to pytket 1.x.

cqc-alec:
After the pytket 2.0 release we should be able to change this to ">= 2.0.0" and remove the handling of, and tests involving, `ClassicalExpBox` in this repo.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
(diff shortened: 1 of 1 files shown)


cqc-melf (approved):

cqc-melf:
Should we do this for the other extensions, too?

cqc-alec:
I've checked all the extensions (I think) and the only ones affected are cutensornet, phir, qir and quantinuum. Have made PRs for all of them.

PR #584, opened 2025-04-23
Add max_cost parameter to process_circuits / submit_program

isobelhooper:
# Description

QuantinuumBackend's `process_circuits` and `submit_program` now accept a `max_cost` parameter that is passed along to the remote Quantinuum systems. (It's ignored for the local emulator.)

# Related issues

(No related issues.)

# Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,12 +2,17 @@
 .. currentmodule:: pytket.extensions.quantinuum
 ```
 
+# Changelog
+
+## Unreleased
+
+- Add `max_cost` parameter to `process_circuits()` and `submit_program()`.
+
 ## 0.46.0 (April 2025)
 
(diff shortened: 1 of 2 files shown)


yao-cqc (changes requested):
Thanks! Could you please add an entry to the changelog?

isobelhooper:
> Thanks! Could you please add an entry to the changelog?

Will do - it looks like the way this has been done in the past for non-release commits is an entry at the top with `## Unreleased` as the header, so I'll do that and fix up where the main changelog title is.

I checked and I don't think the changelog approach is mentioned in the README; is it worth adding some text for that in a separate PR?

yao-cqc (approved):
That looks perfect, thanks!

PR #595, opened 2025-05-29
Improve error message

### LINKED ISSUE #589: Unhelpful error when submitting circuit with classical register exceeding 64 bits
After compiling and then submitting a circuit having a classical register exceeding 64 bits, we get this error:
```
ValueError: classical registers must not have more than 64 bits, you could try to set cut_pytket_register=True in the conversion
```
This is coming from pytket-qir, and isn't very helpful to the user since they can't set that parameter. (Even if they could, it would only split up the temporary registers, not the ones in the original circuit.)

We should catch this case earlier and suggest a workaround such as using `Circuit.rename_units()`.

cqc-alec:
Closes #589 .


cqc-alec:
The test failures seem unrelated (I will investigate separately).

yao-cqc (approved):

PR #600, opened 2025-06-12
fix: replace some old h-series links

CalMacCQ:
# Description

Spotted these old `/h-series` links which go to a 404 page now. I think a redirect used to be in place before the docs site migrated to azure.

It may be worth adding a link checker to the docs build to catch this sort of thing. We do this already for the manual and examples.


yao-cqc (approved):
Thanks!

PR #613, opened 2025-07-08
Update list of op types for validity check; update version and changelog for 0.51.0 release

cqc-alec:
No tests added. Currently these ops aren't implemented for emulators so it's only possible to test on actual devices. I submitted a job using RNG to H1-1 using `process_circuit()`; this passed compilation and is in the queue.

CI failure is unrelated.


cqc-melf (commented):
From my understanding this only works for submission via qasm, but QIR is the default? Should we check that at some point that this requires qasm submission?

cqc-melf (approved):

PR #621, opened 2025-07-22
fix: allow to pass a float as max_cost for QuantinuumBackend

Ariana-Hlavaty:
# Description

Allow to pass a float instead of an int as max cost

# Checklist


cqc-alec:
Done in #622 .

PR #631, opened 2025-08-15
Add tests for RNG and shot number.

### LINKED ISSUE #627: Add tests for RNG and shot number using emulator

cqc-alec:
Closes #627 .

### DIFF
--- a/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -1249,13 +1249,10 @@ class QuantinuumBackend(Backend):
                     quantinuum_circ = circuit_to_qasm_str(
                         c0,
                         header="hqslib1",
-                        maxwidth=(
-                            self.backend_info.misc.get(
-                                "max_classical_register_width", 32
-                            )
-                            if self.backend_info
-                            else 32
-                        ),
+                        # Override the reported value of max_classical_register_width
+                        # (which in August 2025 is 63 for all devices), to enable RNG
+                        # seeding which requires a 64-bit register.
+                        maxwidth=64,
                     )
 
                     used_scratch_regs = _used_scratch_registers(quantinuum_circ)
--- a/tests/integration/backend_test.py
+++ b/tests/integration/backend_test.py
@@ -1580,7 +1580,9 @@ def test_default_pass_serialization() -> None:
 
 @pytest.mark.skipif(skip_remote_tests, reason=REASON)
 @pytest.mark.parametrize(
-    "authenticated_quum_backend_qa", [{"device_name": "H2-1E"}], indirect=True
+    "authenticated_quum_backend_qa",
+    [{"device_name": "H2-1E"}, {"device_name": "H2-1SC"}],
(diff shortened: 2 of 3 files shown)


tests/integration/backend_test.py
```diff
@@ -1576,3 +1576,68 @@ def test_default_pass_serialization() -> None:
...
+    "authenticated_quum_backend_qa", [{"device_name": "H2-1E"}], indirect=True
+)
+def test_shotnum(authenticated_quum_backend_qa: QuantinuumBackend) -> None:
+    circ = Circuit(1).PhasedX(0.5, 0.5, 0).measure_all()
+    shotnum = circ.add_c_register("shotnum", 32)
+    circ.get_job_shot_num(shotnum)
+
+    b = authenticated_quum_backend_qa
+    n_shots = 10
+    h = b.process_circuit(circ, n_shots=n_shots, language=Language.QASM)
```
cqc-melf, line 1635: Should we make the language a parameter of the test, so that we can expand this easily?

tests/integration/backend_test.py
```diff
@@ -1576,3 +1576,68 @@ def test_default_pass_serialization() -> None:
...
+    num1_val = sum(bits[num1_startbit + i] * pow(2, i) for i in range(32))
+    num2_val = sum(bits[num2_startbit + i] * pow(2, i) for i in range(32))
+    assert num0_val <= 5
+    assert num1_val <= 1
+    assert num2_val <= 9
+
+
+@pytest.mark.skipif(skip_remote_tests, reason=REASON)
+@pytest.mark.parametrize(
+    "authenticated_quum_backend_qa", [{"device_name": "H2-1E"}], indirect=True
```
cqc-melf, line 1626: Do we want to run a test for the SC, too?

pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -1249,13 +1249,10 @@ def process_circuits(  # noqa: PLR0912, PLR0915
...
-                            self.backend_info.misc.get(
-                                "max_classical_register_width", 32
-                            )
-                            if self.backend_info
-                            else 32
-                        ),
+                        # Override the reported value of max_classical_register_width
+                        # (which in August 2025 is 63 for all devices), to enable RNG
+                        # seeding which requires a 64-bit register.
+                        maxwidth=64,
```
cqc-melf, line 1255: Do we want to put this in the changelog? Is this something that could potentially cause issues later on?
What happens if the seed register is used later on for something different?

tests/integration/backend_test.py
cqc-alec: I think we can do that when we have QIR submission working.

pytket/extensions/quantinuum/backends/quantinuum.py
cqc-alec: Probably worth putting in the changelog yes. I'll check that we can do other things with the seed register, and if not add a warning.

cqc-melf (approved):

